PR #6, opened 2017-09-13
Fix `skipStories` format in the documentation

sergeyevstifeev:
In the example the spelling is `skip-stories` (kebab-case) but the format is `skipStories` (camel-case).


oblador:
Thanks! 🙇

PR #15, opened 2017-10-02
Replace chrome-headless image by a tagged one

derekstavis:
This PR replaces `armbues/chrome-headless` by `yukinying/chrome-headless:63.0.3209.2`.

The former was having some issues in Docker while starting:

```
Failed to move to new namespace: PID namespaces supported, Network namespace supported, but failed: errno = Operation not permitted
```

Also, the new container is tagged, so a new push will not break new builds.

The only thing I couldn't figure is how to do automatic height adjustment.

### DIFF
--- a/src/targets/chrome/docker.js
+++ b/src/targets/chrome/docker.js
@@ -79,7 +79,7 @@ function createChromeDockerTarget({
       .concat([
         '-p',
         `${port}:${port}`,
-        'armbues/chrome-headless',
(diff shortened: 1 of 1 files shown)


oblador:
Thanks! This seems to break existing visual tests, probably because of the missing height adjustment. Any chance of fixing that? What version did you test? Latest version should have this error fixed I believe.

PR #20, opened 2017-10-21
React native error handling

oblador:
Replaces default React Native crash handler and injects Loki instead to output errors thrown by the story to the terminal instead and automatically recover from the crash by reloading the app. 

As requested by @xaviervia


skellock:
Nice job!

PR #21, opened 2017-10-22
Fix race condition getting the addons channel in react-native

oblador:
...by adding some retry logic. Numbers are arbitrary, please confirm it fixes your issue @Gongreg.


Gongreg:
@oblador, looks like it fixed the issue. 👍

PR #29, opened 2017-10-29
Disable requestAnimationFrame

oblador:
This PR addresses #12 by disabling `requestAnimationFrame` by default via monkey patch.

Also fixes race condition when inserting styles that disables CSS animations and transitions. 

On top of it, adds stories for all these cases. 

Ping @felquis

PR #35, opened 2017-11-12
Run sandboxed storybook bundle in sub process

oblador:
@felquis reported in #29 that loki would sometimes fail with error `TypeError: source.close is not a function`. Upon some investigation the problem was in two parts: 

1. `EventSource` wasn't properly mocked, thus calling `close` on an instance would cause it to crash. 
2. Storybook running in sandboxed VM would includes some retry logic with timers causing the shutdown of the loki process to be slow as it was waiting for all code to execute. 

This PR tries to adress this by running the storybook bundle in a sub process instead and by mocking the `close` method too.

### DIFF
--- a/src/targets/chrome/fetch-storybook-worker.js
+++ b/src/targets/chrome/fetch-storybook-worker.js
@@ -1,0 +1,77 @@
+const vm = require('vm');
+const fetchUrl = require('./fetch-url');
+const debug = require('debug')('loki:chrome:fetchStorybook');
+const getBrowserGlobals = require('./get-browser-globals');
+const { ServerError } = require('../../errors');
+
+async function createStorybookSandbox(baseUrl) {
+  debug(`Fetching iframe HTML and preview bundle JS from ${baseUrl}`);
+  const html = await fetchUrl(`${baseUrl}/iframe.html`);
+  const browser = getBrowserGlobals(html);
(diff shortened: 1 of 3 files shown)


felquis:
Sorry I'm late, I installed loki like this
```
  "devDependencies": {
    "loki": "oblador/loki"
  },
  "loki": {
    "diffingEngine": "looks-same",
    "configurations": {
      "chrome.palm": {
        "target": "chrome.docker",
        "width": 640,
        "height": 480,
        "deviceScaleFactor": 1,
        "mobile": false,
        "fitWindow": false
      }
    }
  }
```

On CI is throws 
```
   
     Fetch list of stories
     Test chrome.palm
     Stop
Timeout
Error: Timeout
    at Timeout._onTimeout (/home/circleci/repo/webapp/node_modules/wait-on/lib/wait-on.js:91:10)
    at ontimeout (timers.js:475:11)
    at tryOnTimeout (timers.js:310:5)
    at Timer.listOnTimeout (timers.js:270:5)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
Exited with code 1
```

Here is [the log](https://circleci.com/gh/pagarme/pilot/734), this time it also does't work on my machine.

PR #37, opened 2017-11-25
Asynchronous stories using `add.async()`

oblador:
For some cases where you need the screenshot to take place at a certain condition, such as your component has loaded some things over the network or has computed a value. 

Not encouraged, but can also be used for regular sleeps. 

Usage: 
```js
storiesOf('MyComponent', module)
  .add('synchronous story', () => <MyComponent />)
  .add.async('asynchronous story', ({ done }) => <MyComponent onDone={done} />);
```


tadashi-aikawa:
> .add.async('asynchronous story', ({ done }) => <MyComponent onDone={done} />);

Can we sleep fixed seconds even if MyComponent doesn't emit `onDone` ?
In my case, MyComponent includes either third-party component or just a word. (they don't have callbacks)

If we can, I'd appreciate it if you could show sample code 🙇

oblador:
@tadashi-aikawa You can do something like: 
```js
.add.async('asynchronous story', ({ done }) => {
  setTimeout(done, 1000);
  return <MyComponent />;
});
```

Might consider adding a shorthand for this, but am afraid it would encourage usage instead of better solutions (when possible).

PR #49, opened 2017-12-26
Fixes hasReactNativeDependency() when the project has no dependencies in package.js

cshell7:
This was an issue for us because we moved what dependencies we had to be peer-dependencies, thus breaking this function.

### DIFF
--- a/src/config/project-package.js
+++ b/src/config/project-package.js
@@ -10,7 +10,7 @@ function getProjectPackage() {
 }
 
 function hasReactNativeDependency(pkg) {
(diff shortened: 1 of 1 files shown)


nuc:
Hi @cshell7,

Thanks for the fix! I experienced exactly the same problem. Can you please append a semi at the end, so the tests pass?

Cheers!

oblador:
@nuc You should be able to push to his branch actually 😊

oblador:
Maybe we should also look in peerDependencies?

PR #52, opened 2018-01-23
Throw if no stories were found

pirelenito:
This most likely is a symptom of something going wrong, and should be considered an error.


oblador:
Thanks!

oblador:
Released in [0.9.1](https://github.com/oblador/loki/releases/tag/v0.9.1)

Issue #51, opened 2018-01-12
 Could not parse url argument "?"


amadt:
I'm getting this in my own app as well as when I run the example app. 

```
loki update v0.7.0
 ❯ Chrome (docker)
   ✔ Start
   ✖ Fetch list of stories
     → Could not parse url argument "?" to pushState against base URL "about:blank".
     Test chrome.laptop
     Stop
Could not parse url argument "?" to pushState against base URL "about:blank".
SecurityError: Could not parse url argument "?" to pushState against base URL "about:blank".
    at HistoryImpl._sharedPushAndReplaceState (/Users/athomson/sandbox/nova/luna/node_modules/jsdom/lib/jsdom/living/window/History-impl.js:73:15)
    at HistoryImpl.pushState (/Users/athomson/sandbox/nova/luna/node_modules/jsdom/lib/jsdom/living/window/History-impl.js:54:10)
    at History.pushState (/Users/athomson/sandbox/nova/luna/node_modules/jsdom/lib/jsdom/living/generated/History.js:89:21)
    at evalmachine.<anonymous>:65329:20
    at Object.dispatch (evalmachine.<anonymous>:50529:7)
    at ConfigApi._renderError (evalmachine.<anonymous>:63464:24)
    at render (evalmachine.<anonymous>:63483:19)
    at ConfigApi.configure (evalmachine.<anonymous>:63498:9)
    at Object.<anonymous> (evalmachine.<anonymous>:57457:22)
    at Object.<anonymous> (evalmachine.<anonymous>:57458:30)
```

Any ideas?

oblador:
Hi, could you share the repo with me so I can investigate?

oblador:
Or oh, it’s the example in this repo? 😄 Think it uses an outdated version of loki by default, let me update it.

amadt:
Yep. I got it in the example repo as well.

ericketts:
I created a fresh react project with `create-react-app`, made a simple button component, added storybook to the app with [`getstorybook`](https://github.com/storybooks/storybook#getting-started), added a story for that button component, and followed the [directions to configure loki](https://loki.js.org/getting-started.html); 

the result is the same error posted above.

lizho-ling:
+1, with latest create-react-app, met this issue.

RobbinHabermehl:
Unfortunately I'm experiencing this issue as well. It concerns a manually configured app, I both tried v0.8.0 and v0.9.0 of loki.

The configuration is very minimal:
```json
[...]

PR #53, opened 2018-01-31
add storybook/ui peer dependency

penx:
as per oblador/loki#51, loki currently throws an error when using storybook 3.2.18 or later.

> Could not parse url argument "?" to pushState against base URL "about:blank".


nuc:
Hey @penx,

the `Could not parse url argument...` error should be gone now. 

Can you try again with the latest version of loki?

PR #57, opened 2018-03-21
Make screenshot timeout configurable

nuc:
Not sure if it's really needed :)

We are just getting some timeouts, and I'd like to increase the duration, just to see if that fixes the problem.

### DIFF
--- a/docs/command-line-arguments.md
+++ b/docs/command-line-arguments.md
@@ -1,10 +1,10 @@
 # Command line arguments
 
-If you run loki via `yarn` or `npm`, you need to make sure to prepend your argument list with `--` so that yarn passes them though to loki. You can also add `./node_modules/.bin` to your `PATH` to be able to run `loki` directly. 
+If you run loki via `yarn` or `npm`, you need to make sure to prepend your argument list with `--` so that yarn passes them though to loki. You can also add `./node_modules/.bin` to your `PATH` to be able to run `loki` directly.
 
 ## `loki test`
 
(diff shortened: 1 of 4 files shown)


nuc:
Hm, seems that `options.chromeScreenshotTimeout` is not available in there. I wonder why 🤔

oblador:
Hey, I think this is a problem when you have too high concurrency. Did you try locally to see if this does indeed fix your problem?

nuc:
@oblador We ended up using the default timeout duration.

Still, we do get some random timeouts, so we thought about adding the [option to retry](https://github.com/verkstedt/loki/pull/1/commits/0a2a035b714cf9a16fa18fee2357f66cdd0ecfd0).

Ideally, we should find the reason of the random failures, but the retries make the builds more likely to pass for now.

Should we create a PR for the retries?

nuc:
Unfortunately, even with the retries, it still fails :(

We will have to dig deeper on this.

oblador:
Want to have retry functionality, but as you said we should fix the real issue. Think it’s chrome itself that doesn’t handle concurrent load that well. Maybe try using a newer version of the docker image?

nuc:
That sounds quite probable. Usually the tests fail on CI, while passing locally. I guess that could be happening due to limited resources on the CI machines.

We will prepare a small PR for the retries and we'll have a look for the concurrency / chrome image.

nuc:
30s is quite reasonable, so closing this PR for now.

PR #64, opened 2018-06-11
Increase shared memory to 1gb

nuc:
That helps to avoid crashes of the docker container, when the parallelism is high.

Should we maybe make it configurable, or would 1gb be ok for now?


nuc:
Merging for now :)

oblador:
Thanks, am a bit slow to review due to vacation. 😊

PR #66, opened 2018-06-17
Fix storiesOf monkey patch

oblador:
Upgrading to a later version of storybook breaks skip/async methods, and this PR fixes that.

### DIFF
--- a/src/targets/decorate-storybook.js
+++ b/src/targets/decorate-storybook.js
@@ -51,8 +51,20 @@ function decorateStorybook(storybook) {
     return stories;
   }
 
-  /* eslint no-param-reassign: ["error", { "props": false }] */
-  storybook.storiesOf = storiesOf;
+  // Monkey patch storiesOf to be able to add async/skip methods
+  const descriptor = Object.getOwnPropertyDescriptor(storybook, 'storiesOf');
+  if (descriptor.writable) {
+    /* eslint no-param-reassign: ["error", { "props": false }] */
+    storybook.storiesOf = storiesOf;
+  } else {
+    Object.defineProperty(storybook, 'storiesOf', {
+      configurable: true,
+      enumerable: true,
+      get: function() {
+        return storiesOf;
+      },
+    });
+  }
 
   function getStorybook() {
     return storybook.getStorybook().map(function(component) {
--- a/src/targets/decorate-storybook.js
+++ b/src/targets/decorate-storybook.js
@@ -57,6 +57,7 @@ function decorateStorybook(storybook) {
     /* eslint no-param-reassign: ["error", { "props": false }] */
     storybook.storiesOf = storiesOf;
   } else {
+    // In recent versions of storybook object this isn't writeable, probably due to babel transpilation changes
     Object.defineProperty(storybook, 'storiesOf', {
       configurable: true,
       enumerable: true,


src/targets/decorate-storybook.js
```diff
@@ -51,8 +51,20 @@ function decorateStorybook(storybook) {
     return stories;
   }
 
-  /* eslint no-param-reassign: ["error", { "props": false }] */
-  storybook.storiesOf = storiesOf;
+  // Monkey patch storiesOf to be able to add async/skip methods
+  const descriptor = Object.getOwnPropertyDescriptor(storybook, 'storiesOf');
+  if (descriptor.writable) {
```
nuc: When the descriptor is not writable? Should we maybe add a small comment to explain?

arthur-feral:
Ah thanks for investigating on this, because, i'm trying right now loki, and i got this problem with 
`stories.add.skip is not a function`

src/targets/decorate-storybook.js
oblador: I'm actually not 100% sure, but assumes it's because of some change to how babel transpiles exports. Want me to add that as a speculation?

nowells:
A future question, what do you think about the approach taken here: https://github.com/tsuyoshiwada/storybook-chrome-screenshot#support-for-adddecorator to support a decorator pattern for specifying screenshots? Would allow you to do the current behavior with a global decorator, or simply targeted story by story, or `storyOf` block level opt-in.

I would be happy to contribute to adding this support, should the be a direction that others are interested in. The current approach is an opt-out only approach, vs supporting an opt-in or opt-out approach depending on your goals.

oblador:
@nowells I've thought about it and even been using decorator approach before, but it had it's own problems. If you want an opt in approach there's the `storiesFilter` config/argument, think that works for you?

nowells:
So this fix breaks `loki` on `@storybook/react@^4.0.0-alpha` with `decorate-storybook.js:63 Uncaught TypeError: Cannot redefine property: storiesOf` for me. This is because we are passing the result of `import '@storybook/react'` through our configureStorybook and that is an ES6 module export which is designed to be immutable and where it's descriptor has `configurable: false` so we cannot redefine it. A hack to get around this would be to just pass `window.__STORYBOOK_CLIENT_API__` into our function instead of the imported storybook/react, but clearly that is a private API. My only other thought is that we can/should switch to the decorator pattern and not try to clobber their API functions (and this might be an example of why it just isn't a good idea). Thoughts?

nowells:
We could also lose some nicer chained syntax of `.add.skip.async` and just use the `setAddon` feature and have that decorate/addParameters that we can query later.

```js
const skippedStories = {};

setAddon({
    addSkip: function(story, storyFn) {
         const kind = this.kind;
         if (!skippedStories[kind]) {
             skippedStories[kind] = [];
         }
         skippedStories[kind].push(story);
         return this.add(story, storyFn);
    }
});
```

And so on for async/skip (we could also move that metadata into addParameters if we really wanted). Obviously this would not let us chain `async` and `skip` and would need `addAsync`, `addSkip`, `addSkipAsync` instead. I might suggest adding `loki` into the name of those functions as well so it is clearer, `skip` and `async` are kind of land grabbing pretty generic concepts.

RishitKedia:
Yup, I can confirm that this fix breaks `loki` with `@storybook/react@^4.0.0-alpha` for me as well. 😭

Just started using `loki` and hit this. Would love to see this get fixed so that I can move along and start using this slick tool. 😋

Oh, and by the way, thanks for all the awesome work you guys have been doing on `loki`! 🤩

nowells:
@oblador thoughts on above? Want me to file a new ticket instead? Would love to discuss best approach going forward. I am happy to code it up once we decide on a preferred direction.

oblador:
@nowells Ideally new ticket yes, primarily for visibility. To address your comments, `__STORYBOOK_CLIENT_API__` actually exists for things like this, but still feels like a hack (well this is too...) and unfortunately doesn't seem to be supported on react-native although I'm sure we can contribute to support it there. 

I agree with you that it's kind of ugly to to monkey patch it as we do now, but I'm also in love with the simplicity of `.skip` that jest has. 

I disregarded decorators as it is limited to runtime execution (unless of course you know of a way), but maybe that's an acceptable trade-off. 

We could also proxy `storiesOf` via the loki package in this manner:
```js
import { storiesOf } from 'loki/react'

storiesOf('MyComponent')
  .add.skip('default', () => <MyComponent />)
```

oblador:
@nowells So I read some code on `setAddon` and yes it does seem like a perfect match actually! Wasn't documented unfortunately, but I'll accept a PR with this functionality 👍

nowells:
@oblador I'll give that setAddon PR a shot. Thoughts on the new names?

opt | skip | asyc | thoughts
--- | ---- | ----- | ------
1️⃣  | addSkip | addAsync | you can switch any existing story from `add` to `addX`
2️⃣  | lokiSkip | lokiAsync | since these are all loki specific, make that clear
3️⃣  | addLokiSkip | addLokiAsync | probably too verbose, but overly specific
4️⃣  | (move to addParameters metadata | (any of the above) | just throwing it out there

I think I prefer 2️⃣  but wanted to get a poll.

oblador:
2 makes sense to me 😊

PR #70, opened 2018-06-19
Allow docker to run from sibling container

nowells:
First of all, thanks for writing `loki`! I am excited to integrate it into my teams toolchain, and looking forward to helping to add features/fixes/etc as they arise.

One of the things that I ran into was that in my team, running `chrome.docker` works great locally, however, we run our Jenkins infrastructure as docker containers (each Jenkins agent is a docker container) so launching `chrome.docker` from Jenkins launches it as a `sibling` container, instead of our process being the host machine and the `chrome.docker` being a docker instance on our host exposing the `port` that we can talk to from `127.0.0.1`.

This PR should fix this problem, allowing the current behavior to work unchanged, however, detecting if the `loki` process is itself running in a docker container, and switching the communication over to use the default `docker bridge network` that is created by default. This default mode lets all docker containers that are part of the default bridge network talk to each other directly via IP address (and port).

Let me know if you would like to see any changes to this PR, or if you want me to solve it another way. I have tested this PR both locally, as well as in our Jenkins-docker setup and both work as expected now.

Thank you for your consideration!

Cheers,
Nowell

### DIFF
--- a/src/targets/chrome/docker.js
+++ b/src/targets/chrome/docker.js
@@ -5,6 +5,7 @@ const { execSync } = require('child_process');
 const execa = require('execa');
 const waitOn = require('wait-on');
 const CDP = require('chrome-remote-interface');
+const fsExtra = require('fs-extra');
 const getRandomPort = require('get-port');
 const { ensureDependencyAvailable } = require('../../dependency-detection');
 const createChromeTarget = require('./create-chrome-target');
@@ -21,11 +22,11 @@ const getLocalIPAddress = () => {
   return ips[0];
 };
 
-const waitOnCDPAvailable = port =>
+const waitOnCDPAvailable = (host, port) =>
   new Promise((resolve, reject) => {
     waitOn(
       {
-        resources: [`tcp:localhost:${port}`],
+        resources: [`tcp:${host}:${port}`],
         delay: 50,
         interval: 100,
         timeout: 5000,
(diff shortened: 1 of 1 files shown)


src/targets/chrome/docker.js
```diff
@@ -40,13 +41,41 @@ const waitOnCDPAvailable = port =>
     );
   });
 
+const getNetworkHost = async dockerId => {
+  let host = '127.0.0.1';
+
+  // https://tuhrig.de/how-to-know-you-are-inside-a-docker-container/
+  const runningInsideDocker = (await fsExtra.exists('/proc/1/cgroup')) &&
```
oblador: Think we can use the sync version of these functions instead. We won't be blocking anything else anyway.

src/targets/chrome/docker.js
```diff
@@ -40,13 +41,41 @@ const waitOnCDPAvailable = port =>
...
+  if (runningInsideDocker) {
+    const { code, stdout } = await execa('docker', [
+      'inspect',
+      '-f',
+      '{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}',
+      dockerId,
+    ]);
+
+    if (code !== 0) {
+      throw new Error('unable to determine ip of docker container');
```
oblador: Sorry for being anal, but can you upper case the first letter and `IP` in the error message?

src/targets/chrome/docker.js
```diff
@@ -96,9 +125,15 @@ function createChromeDockerTarget({
...
-      await waitOnCDPAvailable(port);
-      dockerId = stdout;
-      debug(`Docker started with id ${dockerId}`);
+      try {
+        dockerId = stdout;
+        host = await getNetworkHost(dockerId);
+        await waitOnCDPAvailable(host, port);
+        debug(`Docker started with id ${dockerId}`);
+      } catch (err) {
+        await stop();
```
oblador: Do we have to do `stop()` here or will it be handled down the line regardless?

src/targets/chrome/docker.js
```diff
@@ -5,6 +5,7 @@ const { execSync } = require('child_process');
 const execa = require('execa');
 const waitOn = require('wait-on');
 const CDP = require('chrome-remote-interface');
+const fsExtra = require('fs-extra');
```
oblador: Can you change this import to `fs` just to be consistent?

src/targets/chrome/docker.js
nowells: I was seeing issues in Jenkins of failed processes not being killed, but let me investigate this further and do a separate PR with a targeted bug and fix.

src/targets/chrome/docker.js
nowells: Whoops! I was thinking I should open a different file where it was used to maintain style, but got lazy. Thanks for the callout!

nowells:
> Great job, thanks! Just a few minor comments 

Thanks for the review and help! I will certainly plan on doing more contributions. I started looking at https://github.com/oblador/loki/projects/1 and will certainly try to help where I can.

src/targets/chrome/docker.js
oblador: Sounds good, thx!

PR #75, opened 2018-06-28
Disable will-change CSS property

nuc:
Seems that `will-change` was making some of our tests flaky.

There were some times, 1px differences in some elements, which might randomly happen.

I hope this fixes the issues.

Cheers :)


oblador:
Can you add a test story for this?

nuc:
@oblador Sure! :) Any ideas on how could we test this change?

oblador:
@nuc Just add a story to the web storybook that uses this css property.

nuc:
@oblador better late than never :)

We are also having problems when setting `backface-visibility: hidden` as a performance improvement.

That also causes sometimes minor 1px movements, which break the tests.

Would it make sense to add this as well? 

If not, then we should probably consider allowing the user to provide a custom CSS for overwriting some problematic properties.

oblador:
About the custom CSS stuff this PR is related, https://github.com/oblador/loki/pull/110. Would love to hear your thoughts about it.

oblador:
Released in [0.14.0](https://github.com/oblador/loki/releases/tag/v0.14.0)

PR #78, opened 2018-07-06
Add support to getSelectorBoxSize for complex layouts

techeverri:
## Summary
- Upgrade React, ReactDOM and react-scripts
- Get the box size using querySelectorAll instead of querySelector
- Exclude wrappers from the target
- Add unit tests
- Add visual tests


src/targets/chrome/get-selector-box-size.js
```diff
@@ -1,10 +1,42 @@
 const getSelectorBoxSize = (window, selector) => {
-  const element = window.document.querySelector(selector);
-  if (!element) {
-    throw new Error('Unable to find element');
+  const elements = [...window.document.querySelectorAll(selector)];
+
+  if (elements.length === 0) {
+    throw new Error('Unable to find selector');
```
oblador: Let's keep the error message consistent in case some other part of the codebase looks for it.

src/targets/chrome/get-selector-box-size.js
```diff
@@ -1,10 +1,42 @@
 const getSelectorBoxSize = (window, selector) => {
-  const element = window.document.querySelector(selector);
-  if (!element) {
-    throw new Error('Unable to find element');
+  const elements = [...window.document.querySelectorAll(selector)];
```
oblador: Prefer `Array.from` here

src/targets/chrome/get-selector-box-size.js
```diff
@@ -1,10 +1,42 @@
...
+
+  if (elements.length === 0) {
+    throw new Error('Unable to find selector');
   }
-  const { x, y, width, height } = element.getBoundingClientRect();
-  return { x, y, width, height };
+
+  const removeWrapperElements = element => {
```
oblador: This name is confusing, maybe `isNotWrapperElement` or `doesNotContainOtherElements`?

techeverri: Done! ✅

src/targets/chrome/get-selector-box-size.js
techeverri: What do you mean by consistent? 😕

src/targets/chrome/get-selector-box-size.js
techeverri: This one is faster. 🤓 

https://jsperf.com/array-from-vs-spread-syntax

screen shot 2018-07-06 at 15 59 40

screen shot 2018-07-06 at 15 59 30

screen shot 2018-07-06 at 15 59 10

src/targets/chrome/get-selector-box-size.js
oblador: We need to make sure it's the same message being tested for in [`create-chrome-target.js`](https://github.com/oblador/loki/blob/a455ef88929bba30fa196e40f8a76975bfeb9150/src/targets/chrome/create-chrome-target.js#L149)

src/targets/chrome/get-selector-box-size.js
oblador: Mostly that `Array.from` has wider browser support, but given that we currently only support chrome anyways you can keep it 👍

src/targets/chrome/get-selector-box-size.js
techeverri: Done! ✅

PR #79, opened 2018-07-10
Switch to setAddon to expose loki methods instead of (async, skip)

nowells:
I will leave the old methods (which break in `@storybook/react@4`) but warn when they cannot be created. We should standardize on the new setAddon methods instead going forward.

This fixes a bug when using loki with `@storybook/react@4` which we introduced in https://github.com/oblador/loki/pull/66. It appears that the `storiesOf` is now an  ES6 module export, which cannot be redefined, so our previous approach of clobbering/monkeypatching `storiesOf` no longer works.


oblador:
This is beautiful! Can we add deprecation warnings to consumers of the old API?

nowells:
@oblador not a bad idea! `console.warn` OK?

src/targets/decorate-storybook.js
```diff
@@ -65,8 +65,27 @@ function decorateStorybook(storybook) {
         return storiesOf;
       },
     });
+  } else {
+    console.warn(
```
oblador: Don't think we need to warn about this as this is the new way forward.

oblador:
@nowells Yep! Just make sure to only warn once.

nowells:
@oblador should be all fixed up. Nice feedback!

nowells:
@oblador I assume you will do a version bump and release separately? So just a straight squash merge for now.

oblador:
@nowells Yes we release separately

nowells:
> @nowells Yes we release separately

Thanks for the review!

oblador:
In case you want to do a release:
1. Bump `package.json` version in a separate commit (I do this via github UI :-) ). 
2. Create a new github release with the version as tag (prefixed with `v`, so for example `v1.2.3`)
3. Wait for the CI to publish the package on NPM.

nowells:
@oblador any way you can cut a new release with the recent features.

oblador:
Yup, just want to get this one in: https://github.com/oblador/loki/pull/90

Care to review it?

nowells:
❤️

oblador:
@nowells Released in [0.13.0](https://github.com/oblador/loki/releases/tag/v0.13.0)

PR #87, opened 2018-07-18
Exclude invisible elements from the target

techeverri:
## Summary

- Exclude invisible elements from the target
- Add unit tests
- Add .babelrc to support spread syntax in Jest

## TODO

- Add visual tests


oblador:
Will review more thoroughly later, but I intentionally didn’t transpile the code with babel. Can you remove and refactor to use `Object.assign` instead?

techeverri:
Sure, no problem.

src/targets/chrome/get-selector-box-size.js
```diff
@@ -12,6 +12,27 @@ const getSelectorBoxSize = (window, selector) => {
...
 
+  const isVisisble = element => {
+    let style = {};
+    try {
+      style = window.getComputedStyle(element);
+    } catch (error) {
+      return false;
+    }
+
+    if (
```
oblador: You can just return this if statement instead

src/targets/chrome/get-selector-box-size.js
```diff
@@ -12,6 +12,27 @@ const getSelectorBoxSize = (window, selector) => {
     return !isWrapper;
   };
 
+  const isVisisble = element => {
+    let style = {};
+    try {
+      style = window.getComputedStyle(element);
```
oblador: When would this throw?

src/targets/chrome/get-selector-box-size.js
techeverri: I don't get it 😕

src/targets/chrome/get-selector-box-size.spec.js
```diff
@@ -3,10 +3,21 @@ const getSelectorBoxSize = require('./get-selector-box-size');
...
+          contains: () => element.class === 'wrapper',
+        })
+      ),
+  },
+  getComputedStyle: element => {
+    const { width, height, style = {} } = element;
+    return Object.assign(
+      {},
+      style,
+      { width: `${width}px` },
```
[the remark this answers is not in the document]
techeverri: Done! ✅

src/targets/chrome/get-selector-box-size.js
techeverri: It would throw when `element` is not an instance of `Element`

src/targets/chrome/get-selector-box-size.js
oblador: More like 
```js
return style.visibility === 'hidden' ||
      style.display === 'none' ||
      style.opacity === '0' ||
      style.width === '0px' ||
      style.height === '0px';
```

src/targets/chrome/get-selector-box-size.js
oblador: IDK, feels like that shouldn't happen and that it should really throw in that case.

techeverri:
Not really, I'll resume today =)

src/targets/chrome/get-selector-box-size.js
techeverri: It would need a negation, something like this:

```js
return !(expression);
```

techeverri: Done! ✅

src/targets/chrome/get-selector-box-size.js
techeverri: Done! ✅

oblador:
Cool stuff, will you also add a story to the example project?

techeverri:
Yep, that's in progress =)

PR #89, opened 2018-07-18
Bump prettier and add .prettierrc

### LINKED ISSUE #86: Add Prettier configuration file
Would be awesome to have a `.prettierrc` to use it with code editors or IDEs that support prettier and immediately get the code formatted as soon as you save.

**Prettier / Configuration File**
https://prettier.io/docs/en/configuration.html

The current way is not bad at all, but running `yarn format` manually gets annoying after a while =P

oblador:
Fixes #86


techeverri:
@oblador I'm getting a lot of 
```
error  Replace `'something'` with `"something"`    prettier/prettier
```

techeverri:
`yarn`-ing did the trick 👍

Issue #85, opened 2018-07-17
Failed to fetch stories in CI


geraintf:
Hi,
I'm running the command as per docs to run loki in CI (gitlabs) - `build-storybook && loki --requireReference --reactUri file:./storybook-static`. It builds ok but when it trys to fetch stories it fails because the server is down - I didn't think we had to run storybook if we used static, is that correct or have I misunderstood?

![screen shot 2018-07-17 at 15 47 56](https://user-images.githubusercontent.com/13432246/42825578-36a09036-89da-11e8-8de0-bfe29d0d74b5.png)

oblador:
Hi, looks like it should work; any chance I can have a look at your repo and setup?

nowells:
I believe this is the issue with relative file paths which was recently solved but not released. Try `build-storybook && loki --requireReference --reactUri file:$PWD/storybook-static` (adding `$PWD` to the path)

oblador:
@nowells: Your fix is in 0.12, maybe we need additional work here?

techeverri:
Wasn't it your fix @oblador? 🤔 
Introduced by #81

I think it was not included in v0.12.0

oblador:
That fix was only for Chrome app, docker was already working.

geraintf:
@nowells Adding `$PWD` had the same result

@oblador I'll include as much info as I can below

Files structure:
```
- root
    -ci.yml
    -packages
        - ui-components
            - .loki
            - .storybook (custom config in here)
            - src (ui components in here)
            - stories
            - package.json
```      

  Package.json
```
{
    "engines": {
        "node": ">=8.4.0"
    },
    "scripts": {
        "vr:update": "yarn loki update",
        "vr:test": "yarn loki test",
        "vr:report": "cd .loki && reg-cli ./current/ ./reference/ ./reg-diff/ -R ./report.html -I && cp -r difference/ reg-diff && open ./report.html",
        "vr:ci-test": "build-storybook && loki --requireReference --reactUri file:./storybook-static"
    },
    "devDependencies": {
        "loki": "^0.12.0",
    },
    "loki": {
        "storiesFilter": "^(?:(?!(Spinner)|(Carousel)|(Scenes)).)*$",
        "configurations": {
            "chrome.laptop": {
                "target": "chrome.docker",
                "width": 1200,
                "height": 768
            },
            "chrome.iphone7": {
                "target": "chrome.docker",
                "preset": "iPhone 7"
            }
        }
    }
}
```

[...]

Issue #74, opened 2018-06-27
GitlabCI


frosato-ekino:
First of all, thank you for this great tool, I really like it! 

However, I'm having difficulties to make it run in **GitlabCI**. Target `chome.docker` requires Docker to be installed because it's pulling it own docker image and this does not combine well with GitlabCI.
That means I need to provide to my CI Job a Docker image with Node and Docker installed. 
Since my skills with docker are limited, I was not able to make it work so far. 

```
storybook-test:
  stage: test
  tags: [docker]
  image: a-docker-image/with-node-and-docker
  script:
    - npm i
    - npx build-storybook -s ./public
    - npx loki --requireReference --reactUri file:./public
```
Am I missing something? Do you have any advice to make it work on GitlabCI?

nb: `a-docker-image/with-node-and-docker` is not a real image, this is just ti illustrate the snippet

oblador:
We recently merged a PR improving support for docker in docker (https://github.com/oblador/loki/pull/70), we'll release this to NPM shortly, but meanwhile can you try installing it from github and see if that helps? Also did you look through this one https://docs.gitlab.com/ee/ci/docker/using_docker_images.html?

PR #90, opened 2018-07-18
Add support and documentation for Gitlab CI

oblador:
My docker-fu is not good enough to understand why your fix doesn't work for GitLab @nowells, so I'm resorting to probing for something being available at the `docker` host. Thoughts on how to improve are welcome. 

Should we maintain a docker + node image for people to use?

Related: https://github.com/oblador/loki/issues/85 & https://github.com/oblador/loki/issues/74


src/targets/chrome/docker.js
```diff
@@ -21,6 +21,50 @@ const getLocalIPAddress = () => {
...
+    return true;
+  } catch (error) {
+    return false;
+  }
+};
+
+const getNetworkHost = async dockerId => {
+  if (getIsRunningInsideDocker()) {
+    // Gitlab's docker in docker service is exposed as "docker"
+    if (await getHostExist('docker')) {
```
nowells: Humm, this is going to break my environment because we have a `docker.mycompany-network` server running, and `ping docker` resolves to that, even though I am running a local `docker` daemon that is what I actually want to talk to.

nowells: I don't have a gitlab CI setup myself (but I will try to set one up) to better test/or propose an alternate solution. Let me set about trying to figure that out. (any way we can cut a release before this, since I think this is actually a change that will break us)

oblador: I don’t really like this solution either, but didn’t really manage to find another solution. Thought about using `DOCKER_HOST`, but that’s not right either.

I’mma cut a release for you then.

oblador:
@nowells Created a branch for you: `gitlab.yml` which has a configuration setup and you can find the CI jobs here: https://gitlab.com/trastknast/loki/-/jobs

src/targets/chrome/docker.js
oblador: You can use my branch and setup if you want to save a few minutes :-)

nowells:
OK, I think I added a workable approach to the `gitlab.yml` branch. Ended up that we need to support docker-in-docker (where chrome is launched as a child docker container with our container being the host) and sibling mode where chrome is launched as a sibling container. So I added code for when in docker-in-docker mode you use the hostname of the docker host (if available) to connect to it. Too hacky?

oblador:
I like it, much better than mine 👍 I'll bring your changes into mine

nowells:
Shoot, I finally tested my code in my environment again, and ended up breaking sibling docker support. I need to get better tests setup to exercise this stuff. I'll take a look at fixing it, and adding tests for both modes.

oblador:
Can we add automated tests for this maybe? Will be hairy to work on this stuff otherwise.

oblador:
@nowells Any news on this?

Kristof-Mattei:
I've made, what I think, a simpler version that works off what Docker considers the current context. #181  
When you change `DOCKER_HOST` all will work seamless.

oblador:
Closing due to inactivity.

alanpoulain:
Too bad this PR wasn't merged. How do you want to add automated tests @oblador since it needs GitLab CI to run?
I ended up doing this, which is not really the way GitLab recommends to do it:
```yaml
# .gitlab-ci.yml

stages:
  - test

visual-regression-test:
  variables:
    DOCKER_HOST: localhost
  stage: test
  image: docker:dind
  before_script:
    - dockerd --host=tcp://0.0.0.0:2375 &
  script:
    - apk --no-cache add npm
    - npm ci --prefer-offline
    - npm run test-visual # Insert your test script here
```

PR #91, opened 2018-07-19
Rotate npm token

oblador:


techeverri:
Seems the previous one didn't have the double quotes.
Do we need them now? :thinking:

oblador:
New version of travis cli changed the output perhaps.

Issue #95, opened 2018-07-24
[Suggestion] loki update only unreferenced files


geraintf:
There is a scenario:

I have added a new story that has no reference files. At the same time as adding maybe unintentionally  regress another story, this would now fail `loki test`. 

Before running `loki test` I do `loki update` - this then adds the new reference for my new component but also replaces the old reference for the now broken component meaning I could accidentally introduce regression and not be aware of it.

Suggestion: Add a flag that I can pass into `loki update` that will only add references for those stories which currently have no reference images.

oblador:
I’ll do you one better, add a reference file when there isn’t one when running `loki test`.

geraintf:
Nice! Sounds good, how would that work with the requireReference flag? I guess it'd have to be either or?

oblador:
On CI or with `requireReference` it would not update

geraintf:
Yeah that's what I was thinking 👍

oblador:
@geraintf Have a look at https://github.com/oblador/loki/pull/96

geraintf:
@oblador Just had a play with it and it all seems to work fine. I just had a few thoughts:

- I think the docs should be updated to reflect the new behaviour.
- There's still the chance for someone to accidentally introduce regression by using update.
- The commands I now use for Loki are kinda unclear in terms of the process of how I use it. If I need to  add a new reference image I can run a test or update, but update will also modify existing references, update can also be used to approve changes to files as can the approve command. In my head it makes sense to have commands responsible for a single thing e.g. this is the command you use to add reference images, this is the command you use to test and this is the command you use to approve changes.

geraintf:
@oblador I think your change helps (and removes a potential step from the users workflow). I still think there is a way that a user could unknowingly introduce regression.

- User adds a new story/component 
- User makes a change to some global styles (which breaks another component)
- User does `loki update` to add the new file (they _could_ just do `loki test` but not necessarily) 
- Loki adds the new reference images _and_ approves broken tests.

This leads to potential regression going into production.

oblador:
IDK, this issue applies to jest snapshots too? Also you’d see these files both when committing and making a PR.

What is your proposed alternative workflow?

geraintf:
Yeah, it's one of the downsides of jest, it relies on developers always looking through every failed snapshot (which might not always happen) but I guess the idea should be to make this as difficult as possible to happen.

I guess I imagine it being something like:
- Use `update` to add your initial ref images
- Make some changes and/or add new story
[...]

PR #96, opened 2018-07-24
Auto update missing references

oblador:
Current behaviour is to pass `loki test` locally when a test doesn't have a reference image, but fail it in CI. With this PR `loki test` will still pass, but also add a reference image. Behaviour in CI or with the `--requireReference` flag is unchanged. 

Fixes https://github.com/oblador/loki/issues/95

Bump of jest was needed for promise assertions.

### DIFF
--- a/src/commands/test/test-story.spec.js
+++ b/src/commands/test/test-story.spec.js
@@ -1,0 +1,102 @@
+const fs = require('fs-extra');
+const { getImageDiffer } = require('../../diffing');
+const testStory = require('./test-story');
+
+jest.mock('fs-extra');
+jest.mock('../../diffing');
+
+describe('testStory', () => {
+  const target = { captureScreenshotForStory: jest.fn() };
+  const tolerance = 'mock tolerance';
+  const configuration = 'mock configuration';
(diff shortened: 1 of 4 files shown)


src/commands/test/test-story.spec.js
```diff
@@ -0,0 +1,102 @@
...
+
+  describe('reference image is missing', () => {
+    beforeEach(() => {
+      fs.pathExists.mockReturnValueOnce(Promise.resolve(false));
+    });
+
+    it('throws an error if requireReference option is true', async () => {
+      const options = { requireReference: true };
+
+      return expect(executeWithOptions(options)).rejects.toThrow(
```
techeverri: Nice! 🤠 
Consider adding `expect.assertions(1)` to make sure the assertions was called during a test.

techeverri: More info in the links below:

- Jest / Expect / expect.assertions(number)
https://jestjs.io/docs/en/expect.html#expectassertionsnumber

- An Async Example
https://jestjs.io/docs/en/tutorial-async

Issue #100, opened 2018-08-07
Tab stays open if there is an error when capturing the screenshot


priomsrb:
If an error occurs while capturing screenshots (e.g. the selector can't be found), then the browser tab stays open.

When enough tabs are open, screenshot capturing stops working and timeouts start happening.

PR #101, opened 2018-08-07
Fixes #99 and #100

### LINKED ISSUE #99: Timeout happens when selector has zero width or height
When a story has blank content, `Page.captureScreenshot()` waits indefinitely causing a timeout. Instead, an error should be thrown immediately.

priomsrb:
Fixes #99 and #100

### DIFF
--- a/src/targets/chrome/create-chrome-target.js
+++ b/src/targets/chrome/create-chrome-target.js
@@ -167,6 +167,12 @@ function createChromeTarget(
           debug(`Getting viewport position of "${selector}"`);
           const position = await getPositionInViewport(selector);
 
+          if (position.width === 0 || position.height === 0) {
+            throw new Error(
+              `Selector "${selector} has zero width or height. Can't capture screenshot.`
+            );
+          }
+
           debug('Capturing screenshot');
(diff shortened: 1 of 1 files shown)


src/targets/chrome/create-chrome-target.js
```diff
@@ -237,17 +243,18 @@ function createChromeTarget(
     const tab = await launchNewTab(tabOptions);
     try {
       await withTimeout(options.chromeLoadTimeout)(tab.loadUrl(url));
+      const screenshot = await tab.captureScreenshot(selector);
+      await fs.outputFile(outputPath, screenshot);
+      return screenshot;
```
oblador: I'm not certain about how JS will treat this, but will the `finally` block be executed even after a `return` statement?

techeverri:
It seems the longer is complaining about a block without return statement

src/targets/chrome/create-chrome-target.js
priomsrb: I think finally should always get executed. Still, I'll move the return statement for clarity.

priomsrb:
No problem. Thanks for the quick review :)

PR #102, opened 2018-08-08
Be more graceful when the selector is not valid: take a screenshot of 'body' in that case.

priomsrb:
Sometimes my stories look fine in the browser, but Loki complains that the selector isn't valid.

This can happen in many cases. For example: if my component is absolutely positioned, then the selector would have a 0 width and height. In those cases it might be better to just capture a screenshot of the entire body rather than throw an error.

One of the things I like about Loki is that it requires minimal configuration and tweaking to get it working. So I think it should avoid throwing errors if the story looks fine in the user's browser.

### DIFF
--- a/src/targets/chrome/create-chrome-target.js
+++ b/src/targets/chrome/create-chrome-target.js
@@ -148,24 +148,22 @@ function createChromeTarget(
       await executeFunctionWithWindow(awaitLokiReady);
     };
 
-    const getPositionInViewport = async selector => {
-      try {
-        return await executeFunctionWithWindow(getSelectorBoxSize, selector);
-      } catch (error) {
-        if (error.message === 'Unable to find selector') {
-          throw new Error(
-            `Unable to get position of selector "${selector}". Review the \`chromeSelector\` option and make sure your story doesn't crash.`
-          );
-        }
-        throw error;
-      }
-    };
+    const getPositionInViewport = async selector =>
+      executeFunctionWithWindow(getSelectorBoxSize, selector);
 
     client.captureScreenshot = withRetries(options.chromeRetries)(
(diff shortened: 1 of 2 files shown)


oblador:
Thanks! This looks good to me, @techeverri WDYT?

techeverri:
Taking the screenshot of the whole body, Isn't this going to include the Storybook elements? :thinking:

Seems this could give people the false sense that everything is fine even when they're expecting the screenshot to be a portion of the screen (e.g. using a decorator). IMO returning false positives is not a good thing, it makes hard to distinguish between an error and the fallback.

priomsrb:
Hey @techeverri . Sorry for the delayed reply.

> Taking the screenshot of the whole body, Isn't this going to include the Storybook elements? 🤔

Loki takes screenshots of the `/iframe.html` page which doesn't include storybook elements. So taking a screenshot of the body will not include storybook elements.

> Seems this could give people the false sense that everything is fine even when they're expecting the screenshot to be a portion of the screen (e.g. using a decorator).

Could you clarify what you meant by "using a decorator"? I didn't fully understand this part.

> IMO returning false positives is not a good thing, it makes hard to distinguish between an error and the fallback.

I'm not sure I would call it a false positive. Rather, in the current situation, viewing certain stories in the browser works fine, but when loki tries to take a screenshot it fails. I'll link some examples where I've encountered this to clarify what I mean.

priomsrb:
Here's a link to 2 examples that currently cause an error: https://github.com/oblador/loki/issues/115#issuecomment-442280516

The changes in this PR will fix the errors by taking a screenshot of the body.

oblador:
With recent developments in this project, stuff like fixed/absolute positioning is now better supported, so I'm going to close this PR.

PR #108, opened 2018-10-23
Docker ext target

jacksbox:
Adds an external started docker container as a new target.

Configuration option `target: 'chrome.dockerExt'`
Cli Params`--dockerHost=xxx --dockerPort=xxx`


oblador:
Why did you close the PR?

nuc:
@oblador Oops, that was done accidentally :) We're now cleaning up our changes and we will prepare a new PR for that.

Issue #109, opened 2018-10-25
Detect loki tests from the app


vvscode:
Is it possible or not? (Or feature request, if not) - add class / detect by user-agent and so on.

What for it might be used - to ignore/disable particular areas on screen shot ( set opacity for example )

techeverri:
> What for it might be used - to ignore/disable particular areas on screen shot ( set opacity for example )

Elements with the following properties would be identified not visible

https://github.com/oblador/loki/blob/f8954830b86b1308c45e607046b9f1c3cef455bf/src/targets/chrome/get-selector-box-size.js#L15-L25

vvscode:
I realize it. But currently, I have components with dynamic content. So it should be out of screenshot. Only for tests.

in `gemini` for example, I could set ignore selectors. In loki I see no such options. So I want ignore valuable elements, but only for tests

Is it possible?

vvscode:
https://gemini-testing.github.io/doc/tests.html 

```ignoreElements('.selector1', {every: '.selector2'}, ...) — elements, matching specified selectors will be ignored when comparing images.

.selector1 — Ignore only the first matched element.
{every: '.selector2'} — Ignore all matched elements.```

**Comment by vvscode** [comment]
What do I mean - for example at https://github.com/oblador/loki/blob/02301c9548c68ca3e60c5033ce23c1d19d341d3d/src/targets/chrome/create-chrome-target.js#L246 add `loki-tests` attr to `html` tag so in my styles I could write

```css
/* loki specific styles */
[loki-tests] .map {
    opacity: 0;
}
```

vvscode:
I'm ready try implement it by myself if the team is ok about this suggestion

oblador:
It's actually part of the roadmap :-) https://github.com/oblador/loki/projects/1#card-5208258

Would love some help if you need it soon.

vvscode:
@vvscode @oblador  it would be nice to see some contributors guide. Not obvious how to test changes and so on - I spent 3 hours for small update

vvscode:
and this one might be closed, I guess =)

PR #110, opened 2018-10-26
Mark root element with attr

vvscode:
https://github.com/oblador/loki/issues/109

https://github.com/oblador/loki/projects/1#card-5208258

With this update, I'm able to use  next code in styles 

```css
  *[loki-test] .loki-invisible {
    opacity: 0;
  }
```
or from js:
```javascript
if (window._isLokiTest) {
    // loki specific logic
}
```

### DIFF
--- a/src/targets/chrome/add-loki-session-marker.js
+++ b/src/targets/chrome/add-loki-session-marker.js
@@ -1,4 +1,7 @@
-const addLokiSessionMarker = window =>
+const addLokiSessionMarker = window => {
   window.document.querySelector(':root').setAttribute('loki-test', true);
+  /* eslint-disable no-underscore-dangle, no-param-reassign */
+  window._isLokiTest = true;
+};
 
 module.exports = addLokiSessionMarker;
--- a/src/targets/chrome/add-loki-session-marker.js
+++ b/src/targets/chrome/add-loki-session-marker.js
@@ -1,0 +1,4 @@
+const addLokiSessionMarker = window =>
+  window.document.querySelector(':root').setAttribute('loki-tests', true);
+
+module.exports = addLokiSessionMarker;
--- a/src/targets/chrome/add-loki-session-marker.js
+++ b/src/targets/chrome/add-loki-session-marker.js
@@ -1,4 +1,4 @@
 const addLokiSessionMarker = window =>
-  window.document.querySelector(':root').setAttribute('loki-tests', true);
+  window.document.querySelector(':root').setAttribute('loki-test', true);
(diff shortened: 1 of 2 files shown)


src/targets/chrome/add-loki-session-marker.js
```diff
@@ -0,0 +1,4 @@
+const addLokiSessionMarker = window =>
+  window.document.querySelector(':root').setAttribute('loki-tests', true);
```
oblador:
```suggestion
  window.document.querySelector(':root').setAttribute('loki-test', true);
```

Somehow feels this should be in singularis.

oblador:
Should we maybe expose a function that one could call to test this procedurally? Like `isLokiTest()`. 

Also should we support this on native perhaps?

src/targets/chrome/add-loki-session-marker.js
vvscode: updated

vvscode:
@oblador I have no experience in using Loki with native apps. As far as I know ( from my experience in RN) - there it's easier to implement with `isLokiTest` function. However, I have a vision of implementing it for web, not for native 

Are you going share contributors guide?

oblador:
If you're OK with doing a `isLokiTest()` implementation for web I can do one for native in a separate PR.

vvscode:
@oblador  see no point for function. just have `window._isLokiTest` field is enough for me

vvscode:
@oblador what do you think?

vvscode:
Yep. But importing the module in tests make impossible import that module out of tests. Just because loki-tests is something aside from the main application. I understand your point of view, but don't understand how it's possible to implement

oblador:
Hmm, I wouldn't want to litter my components with test code, but rather use it in the story itself to pass a prop. Regardless, importing a function or checking a global would be just as bad to do in your production code? If you want to be able to remove it with a minifier etc, then I guess some thing like `process.env.LOKI_TEST` would be better no?

vvscode:
but it also pollution of global variable, isn't it? And more than that - you have to care about passing that prop to the components

vvscode:
and in web it should be `window.preocess.env.LOKI_TEST`?

oblador:
`process.env.LOKI_TEST` is pollution yes, but at least there's tooling to have it replaced with the primitive value and thus tree-shakeable. 

Passing it as a prop in stories is to me fine because stories is your test code and that's where it belongs IMHO. You could also have a context provider for all your stories with some generic theme settings.

PR #111, opened 2018-10-27
Add marker for native

vvscode:


wakeless:
Out of interest, what's the purpose for this change? How are you using it in your code?

vvscode:
Not for RN - just for React in Web ( to set opacity for non-valuable elements )

PR #118, opened 2018-12-09
Maintain access to resolveAssetSource after monkey-patch

wakeless:
Currently using `Image.resolveAssetSource` is not possible as the ready state emitting Image monkey patch does not re-export it.

image

This resolves this issue.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "fs-extra": "^7.0.1",
     "get-port": "^4.0.0",
     "gm": "^1.23.1",
+    "hoist-non-react-statics": "^3.2.1",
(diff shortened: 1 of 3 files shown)


oblador:
Thanks! What do you think about using something like https://github.com/mridgway/hoist-non-react-statics instead?

wakeless:
@oblador That's now sorted. Introduces a dependency of `react-is` but I think that's probably ok.

wakeless:
@oblador any chance I could get this wrapped up into a release?

oblador:
@wakeless Done!

wakeless:
@oblador 💓 thanks a lot!

PR #122, opened 2019-01-03
Remove the StatusBar controls and force the implementor to do it at Storybook mount time

wakeless:
I have had issues with the StatusBar re-appearing. 

There appears to be a half-implemented status bar control (the cache is never set). I think the easiest approach is to remove this, and allow the controls of it where you invoke `<Storybook />`

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -10,4 +10,5 @@
   * [loki init](command-line-arguments.md#loki-init)
 * [Continuous integration](continuous-integration.md)
 * [Handling flaky tests](flaky-tests.md)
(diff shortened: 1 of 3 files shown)


oblador:
Thanks for your PR! Are you sure this fixes it, couldn't it be one of your components setting the status bar? I'd like loki to do as much as possible automatically of configuring stuff like this. Maybe we should be hiding the status bar for each story before taking the screenshot

wakeless:
It's a weird one, you are right, one of our components was causing the issues, and then there was an issue with it not setting itself back to how it was (rather than just back on)

PR #127, opened 2019-01-15
Allow configuration of diffing libraries directly from JSON

wakeless:
I need to use a number of the looks-same configuration to stamp out flakeys in our build. This allows me to configure the looks-same library directly.

It should also respect the chromeTolerance config as a fallback.

To configure you will add to the json:

```
"looksSame": {
  "ignoreAntialiasing": true
}
```


wakeless:
@oblador Will fill this out with documentation etc if you are happy with the general approach.

wakeless:
@oblador Thanks. Have added a test for GM, and added docs. I spent a heap of time trying to test the looks-same implementation, but couldn't make stubbing work too well so have skipped it.

mriddle:
Was there anything else we needed to do to get this in? I'm keen for this change 🙌

Issue #94, opened 2018-07-24
Timeout fail for Chrome on CircleCi


adekbadek:
Hi, I have this error on CircleCI (locally everything is fine):
```
loki test v0.13.0
[14:48:46] Chrome (docker) [started]
[14:48:46] Prepare environment [started]
[14:48:52] Prepare environment [completed]
[14:48:52] Start [started]
[14:49:25] Start [failed]
[14:49:25] → Timeout
[14:49:25] Chrome (docker) [failed]
[14:49:25] → Timeout
Timeout
Error: Timeout
```

I increased the timeout in [`docker.js`](https://github.com/oblador/loki/blob/7af2feb16f8cf01ff711c654636846d832fa8cbd/src/targets/chrome/docker.js#L31) to `30000` as suggested in #38 but this did not fix the issue.

CircleCI is using `circleci/ruby:2.5-node` image.

Any tips on debugging this issue?

oblador:
Think this is related to #92 and #90

jagregory:
I'm seeing the same issue with CircleCI. Could it be related to the way Docker networking is setup in Circle? https://circleci.com/docs/2.0/building-docker-images/#separation-of-environments That sounds suspiciously like Loki wouldn't be able to connect directly to the Chrome containers, right?

dbartholomae:
I actually get the following error message:
```
[01:29:21] → Command failed: docker images -q yukinying/chrome-headless:63.0.3230.2
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```
Will investigate more.

dbartholomae:
@oblador My naive approach would be to set up an additional target that doesn't spin up docker, but instead connects to a predefined docker container. This would require users that set up CircleCI to ensure the correct container is running and is reachable, but to be honest I don't see a good way to spin up the docker container from inside a CircleCI command. I'll experiment with this and hopefully provide a PR :)

oblador:
@dbartholomae Looking forward to it!

dbartholomae:
@oblador I've played around with some concepts and in the end came up with the following thoughts:
Instead of having a separate target the original target needs to be able to switch over to using an already running Docker container. Otherwise the user of loki would need to run different target configs on CI and local, which is bound to lead some false positives in the tests.
Here's my current code:
https://github.com/dbartholomae/loki/tree/add_remote_chrome_target

[...]

PR #129, opened 2019-01-21
Add remote docker chrome target for CI server

MaciejJaroszewski:
PR resolves:

1. Problem with connect to external docker image with chrome-headless
2. Problem with navigating through different stories via iframes - probably bumping dependencies, additional configuration serve-handler and increasing timeout would help
3. process exits when all test passes - after fix first and second problem, all tests would pass if correct, but proces didn't exit and circle CI waiting for 10 minutes and after that CI server kill process and throw error (circleCI timeout)

Thanks @dbartholomae for initial commits.

Resolve issue #94


koshov:
@MaciejJaroszewski thank you for pushing this forward!
I tested your fork on one of my projects, but am still getting the Timeout on CircleCI. Have you been able to revisit this? I would be happy to lend a hand if there are any details on what is outstanding to get Loki working on CircleCI.

codrin-iftimie:
I know I'm late to the party, but would it be easier to use the `"chrome.app"` as the target in this case?
And have the `loki.config.js` file select which one should be used as target based on a ENV variable, like `CI`
Something like:
```js
// loki.config.js

module.exports = {
  configurations: {
    desktop: {
      target: process.env.CI ? "chrome.app" : "chrome.docker",
      width: 1366,
      height: 768,
      deviceScaleFactor: 1,
      mobile: false,
    },
  },
};
```

kadosa:
I know this PR is quite old, but I'm wondering if there's a chance to get this looked at again? Or if there's a suggested way of using `chrome.docker` on CircleCI?

Issue #113, opened 2018-10-28
Add more info about `chromeTolerance`


vvscode:
I was looking about more info according to `chromeTolerance` and couldn't. At least range of possible values? Are we talking about 1-2-3-10 or about 0.01-0.2 and so on?

And please share reference for comparison option 

Thanks in advance

oblador:
Unfortunately this depends on the diffing tool used. I'd say it's more in the `0.01-0.2` range, but I generally recommend setting it to `0` unless you're having a good reason not to.

mowbell:
@oblador Why reason currently does not passed pass --chromeTolerance option on react native simulators and emulators?
I think is useful for emulators too, due that flags are runned by diffengine (gm-looks-same), and it does not depend of target (chrome simulator, etc)
Currently on **src/commands/test/run-tests.js** createTargetTask does not pass any tolerance flag
```js
const createTargetTask = configurations => {
    const { target } = configurations[Object.keys(configurations)[0]];
    switch (target) {
      case 'chrome.app': {
        return getTargetTasks(
          'Chrome (app)',
          createChromeAppTarget({
            baseUrl: options.reactUri,
            chromeFlags: options.chromeFlags,
          }),
          configurations,
          options.chromeConcurrency,
          options.chromeTolerance
        );
      }
      case 'chrome.docker': {
        return getTargetTasks(
          'Chrome (docker)',
          createChromeDockerTarget({
            baseUrl: options.reactUri,
            chromeDockerImage: options.chromeDockerImage,
            chromeFlags: options.chromeFlags,
          }),
          configurations,
          options.chromeConcurrency,
          options.chromeTolerance
        );
      }
      case 'ios.simulator': {
        return getTargetTasks(
          'iOS Simulator',
          createIOSSimulatorTarget(options.reactNativeUri),
          configurations,
          //options.chromeTolerance <--- IT's NOT PASSED
        );
      }
      case 'android.emulator': {
        return getTargetTasks(
          'Android Emulator',
          createAndroidEmulatorTarget(options.reactNativeUri),
          configurations,
          //options.chromeTolerance <--- IT's NOT PASSED
        );
      }
      default: {
        return die(`Unknown target "${target}`);
      }
    }
  };
```

levymetal:
@abejfehr you can pass options directly to the diffing engine by using the name of the engine as a key in your config. For example:

```json
{
  "diffingEngine": "gm",
  "gm": {
    "tolerance": 0.00024
  }
}
```

[...]

PR #132, opened 2019-03-15
Rename `chromeTolerance` to `tolerance` and pass to all targets

mriddle:
What
-----

**This is a breaking change. When users update they'll need to specify `tolerance` instead of `chromeTolerance`**

`tolerance` is respected by `gm` and `looks-same` yet we weren't passing the option for the iOS simulator and Android emulator. This patch rectifies that.

I've removed the `tolerance / 100` in `diffing/gm` for consistency between the two engines so that the may user decide. 

Lastly, I've added a section on tolerance to clear up any ambiguity which should resolve https://github.com/oblador/loki/issues/113

If you're troubleshooting images and want to see the numeric diff between your values you can do so via the command line

```
gm compare -metric MSE  reference-image.png compare-image.png

Image Difference (MeanSquaredError):
           Normalized    Absolute
          ============  ==========
     Red: 0.0000153787        1.0
   Green: 0.0000153787        1.0
    Blue: 0.0000153787        1.0
 Opacity: 0.0000000000        0.0
   Total: 0.0000115340        0.8
```

The `total` is compared with the `tolerance` set


mriddle:
🤔the build is failing with the error below but I can't see how my changes have impacted it nor can I get it to fail locally.

```
[01:38:55] Asynchronous render: Logo with 1s delay [failed]
[01:38:55] → Screenshot differs from reference, see .loki/difference/chrome_iphone7_Asynchronous_render_Logo_with_1s_delay.png
```

src/commands/test/parse-options.js
```diff
@@ -27,7 +27,7 @@ function parseOptions(args, config) {
     chromeLoadTimeout: parseInt($('chromeLoadTimeout'), 10),
     chromeRetries: parseInt($('chromeRetries'), 10),
     chromeSelector: $('chromeSelector'),
-    chromeTolerance: parseFloat($('chromeTolerance'), 10),
+    tolerance: parseFloat($('tolerance')),
```
mriddle: Removed the `10`, assuming it was there accidentally. `parseInt` takes a `radix` parameter. `parseFloat` does not 😸

mriddle:
The failures seem to be isolated to Travis, https://github.com/oblador/loki/pull/127 has failed as well with the same errors.

oblador:
Hi @mriddle. I'm thinking of removing tolerance completely as these tests should be deterministic. The reason I have it at all is since some users might use the `chrome.app` target which is not deterministic. Why you need it?

Another thing I don't like about it is that the tolerance input is interpreted quite differently, the division by 100 is an effort to normalize them somewhat. So I'd prefer to actually be _more_ specific in this flag.

mriddle:
Hi @oblador, the tolerance is being used to account for variations between different systems. I've noticed that when comparing a reference image I've taken locally to the current image taken on CI there's the smallest amount of variance causing our specs to fail. 

The images look exactly the same, same font, same colours, etc, could be something like anti-aliasing coming into play. The variance is so minor I'd rather set a tolerance than sink a bunch of time into investigating the cause.

priomsrb:
For what it's worth, I've encountered a similar issue when using a 3rd party UI library. There was one component that would have a (1,1,1) rgb value difference in different runs. After investigating it for a few hours, I couldn't figure out the cause. So I had to use the tolerance option to remove the flakiness from my test.

oblador:
@mriddle Do you think https://github.com/oblador/loki/pull/127 would serve your needs?

mriddle:
@oblador yes, with the changes in #127 I can pass tolerance to the diffing engines. I believe these changes are still beneficial as it brings some consistency to 'tolerance' but I'll leave that decision to you.

PR #139, opened 2019-04-09
Add option to run docker commands with sudo

ldabiralai:
Unfortunately in our Jenkins CI we are restricted an only able to run `docker` commands with `sudo`, because of this loki is unable to run with the `chrome.docker` target.

This PR adds a `dockerAsSudo` option to resolve this, it'll run all `docker` commands with `sudo` prepended if the option has been specified.

More than happy to take feedback & answer any questions!

### DIFF
--- a/docs/command-line-arguments.md
+++ b/docs/command-line-arguments.md
@@ -36,6 +36,7 @@ yarn loki test -- --port 9009
 | **`--targetFilter`**           | Regular expression for targets that should be tested.                                                                                           | _None_                                       |
 | **`--requireReference`**       | Fail stories without reference image, useful for CI.                                                                                            | _False, true for CI_                         |
 | **`--verboseRenderer`**        | Plain text renderer, useful for CI.                                                                                                             | _False, true for CI_                         |
+| **`--dockerWithSudo`**        | Run docker commands with sudo                                                                                                             | `false`                         |
 
 ## `loki update`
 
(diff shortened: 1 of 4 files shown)


src/targets/chrome/docker.js
```diff
@@ -180,7 +190,8 @@ function createChromeDockerTarget({
 
   process.on('SIGINT', () => {
     if (dockerId) {
-      execSync(`${dockerPath} kill ${dockerId}`);
+      const maybeSudo = dockerAsSudo ? 'sudo' : '';
+      execSync(`${maybeSudo}${dockerPath} docker kill ${dockerId}`);
```
oblador: This looks wrong to me. Shouldn't it be `execSync(`${maybeSudo} ${dockerPath} kill ${dockerId}`);`?

docs/command-line-arguments.md
```diff
@@ -36,6 +36,7 @@ yarn loki test -- --port 9009
 | **`--targetFilter`**           | Regular expression for targets that should be tested.                                                                                           | _None_                                       |
 | **`--requireReference`**       | Fail stories without reference image, useful for CI.                                                                                            | _False, true for CI_                         |
 | **`--verboseRenderer`**        | Plain text renderer, useful for CI.                                                                                                             | _False, true for CI_                         |
+| **`--dockerWithSudo`**        | Run docker commands with sudo                                                                                                             | `false`                         |
```
oblador: Seems incorrect, it's `dockerAsSudo` in the code base. I think I prefer either `dockerWithSudo` or `dockerAsSuperUser`, WDYT?

ldabiralai:
Thanks @oblador! Both good points, I've updated the PR to resolve these 🙂

Issue #92, opened 2018-07-20
Gitlabs CI timeout on debugger


geraintf:
I am getting a timeout error when trying to run loki in gitlabs CI.

`Timed out waiting for Chrome Debugger to appear on  172.x.x.x:42676`

Using this version of Loki: https://github.com/oblador/loki/pull/90

.gitlab-ci.yml
```
visual-regression-test:
  <<: *apply_runner_tags
  <<: *skip_on_tags
  stage: test
  image: an-internal- (this image installs node 8.x and installs docker via apk add --update --no-cache docker)
  script:
    - cd packages/ui-components
    - docker info
    - npm run vr:ci-test (build-storybook && loki --requireReference --reactUri file:./storybook-static)
```

loki config
```
  "loki": {
    "storiesFilter": "^(?:(?!(Spinner)|(Carousel)|(Scenes)).)*$",
    "configurations": {
      "chrome.laptop": {
        "target": "chrome.docker",
        "width": 1200,
        "height": 768
      },
      "chrome.iphone7": {
        "target": "chrome.docker",
        "preset": "iPhone 7"
      }
    }
  }
```

![screen shot 2018-07-20 at 16 55 21](https://user-images.githubusercontent.com/13432246/43012624-86b48b52-8c3e-11e8-93d1-475c0e79e48d.png)

I believe it comes from this line https://github.com/oblador/loki/pull/90/files#diff-b75c1e1e6b0fa125e76a9f03fe7716c5R171, it seems that maybe to it thinks its found docker (or maybe it has) but the ip/port is wrong.

geraintf:
After debugging it seems it is due to docker sibling being broken as part of #90 - I guess I'll have to wait until its fixed as part of this

oblador:
Any progress on that one @nowells?

nowells:
Not yet, I'll pick it back up tonight and see what I can do. Sorry for the delay!

geraintf:
Hi, Is there any update on this? 
Thanks

geraintf:
I'm still having trouble using loki as part of gitlabs until this gets resolved (and this means that we don't get much value from it until then). I would offer to assist with the PR but docker isn't really my forte

PR #14, opened 2017-10-02
ci: skip seccomp on ci

felquis:
closes #13 

With this change seccomp files are skipped from CI, but in your local environment seccomp will be used.

PR #150, opened 2019-06-03
Make docker seccomp usage opt-out

oblador:
For me and potentially others (#92, #94) Chrome in Docker wouldn't work on CI without seccomp settings. 

This PR reverts #14 and adds a `--chromeDockerWithoutSeccomp` flag to explicitly opt out of it. **It is  breaking change!**

@ldabiralai Do you think we could get rid of the `dockerAsSudo` flag now?


ldabiralai:
@oblador I'll test if this works for us instead of `dockerAsSudo` :)

oblador:
Amazing, do you think you could test without a release?

ldabiralai:
Yeah, that should be fine!

ldabiralai:
@oblador unfortunately running with `chromeDockerWithoutSeccomp` and not `dockerWithSudo` doesn't work for us. We have the same error as before.

oblador:
Thanks for reporting, then we'll keep it!

PR #160, opened 2019-07-10
Fix issue with lack of execute

szymansd:
During: https://github.com/oblador/loki/commit/d5215172d2da54a8774b25094281d120a7f77cd8#diff-b75c1e1e6b0fa125e76a9f03fe7716c5R167 an error has been made and `execute` is not passed anymore to `getNetworHost` function which causes an error like below:

Screenshot 2019-07-10 at 13 25 11

This PR should fix this issue.


sweiler:
I have the same issue, only fix for us currently is to go back to 0.15.1.

PR #163, opened 2019-07-25
Adding Image.propTypes to ready-state-emitting-image to work with react-native-video

shanez:
react-native-video sets up it's propTypes by using `Image.propTypes.resizeMode`. 

`ready-state-emitting-image` does not have propTypes specified, so this causes `react-native-video` to crash when used with loki.

This PR fixes the issue mentioned by adding propTypes onto the `ready-state-emitting-image`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "loki",
-  "version": "0.17.0",
+  "version": "0.17.1",
   "description": "Visual Regression Testing for react-storybook",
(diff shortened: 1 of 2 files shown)


oblador:
I'm not sure why hoist statics wouldn't include this, but added it anyways. Wasn't able to push to your branch so I fixed it and cherry-picked to master directly instead. Thanks!

PR #165, opened 2019-08-07
Add outputFilenamesLowercase option

### LINKED ISSUE #164: Allow configuring output files to be lowercase
OS X is one such operating system where files are case-insensitive. This has caused a problem or two when a story name has had character cases changed. Changing case of a story is acceptable, since stories are just code, not file names, but since file names are being generated from them, it would be ideal if one could configure the generated file names to be all lowercase to avoid changing the case of the file (creating a second file instead of overwriting the existing).

r33beers:
`outputFilenamesLowercase` option, when set to true, forces all output filenames to be lowercase. This is useful for operating systems with case insensitive file systems.

Fixes #164


oblador:
I think a better approach would be to allow fully customizable filenames by passing a function or similar.

r33beers:
Can you elaborate more on what is driving the need for a fully customizable solution? I was going for something simple that solves #164.

oblador:
This has been superseded by #213

PR #166, opened 2019-08-08
Retry all docker createNewDebuggerInstance calls with backup

oblador:
Fixes `socket hang up` in some cases.


oblador:
Doesn't really, so closing

PR #168, opened 2019-08-20
Pass missing execute to getNetworkHost

sweiler:
The signature of getNetworkHost changed, but the only call to it was not adapted.


sweiler:
Sorry.. Didn't see the other PR. So this is a duplicate of #160.

PR #170, opened 2019-09-13
add `chrome.autoExpandHeight` option

tjenkinson:
which when enabled will extend the viewport height if necessary so that the area that is being screenshotted does not go below the fold.

With chrome 76 (and probably earlier) any area below the fold shows up as white.

It looks like a while back the screenshot would still work even when the area went below the fold, because with the default 'yukinying/chrome-headless' image (that hasn't been updated in a while) anything below the fold is still captured.

Given that with an older version of chrome anything below the fold would be captured, I am wondering if this should be considered a bug fix and `autoExpandHeight` default to `true` instead of `false`?


oblador:
I can see scenarios where this won’t work perfectly, but this seems like a good improvement!

oblador:
I think this is probably expected behavior, so lets make it default and make the option to disable instead of enable, wdyt?

oblador:
I think travis broke something in an upgrade, need to fix my setup somehow.

oblador:
If we have it enabled by default I'd prefer to have the option to disable it, basically rename it to `disableAutomaticViewportHeight`, `disableDynamicViewportHeight` or similar.

oblador:
Good point, totally forgot about that. Did you give it a try to see if it fixes your use case?

oblador:
Travis failures should be fixed now if you rebase on master.

oblador:
Very possible, if so would you mind removing it or augment the setting to your new implementation?

docs/configuration.md
```diff
@@ -82,17 +82,17 @@ Opposite to `skipStories`.
...
+| **`preset`**                         | _string_  | Predefined bundled configuration, possible values are `Retina Macbook Pro 15`, `iPhone 7`, `iPhone 5`, `Google Pixel`, `A4 Paper`, and `US Letter Paper`. | `chrome.*`   |
+| **`userAgent`**                      | _string_  | Custom user agent.                                                                                                                                        | `chrome.*`   |
+| **`width`**                          | _integer_ | Browser viewport width.                                                                                                                                   | `chrome.*`   |
+| **`height`**                         | _integer_ | Browser viewport height.                                                                                                                                  | `chrome.*`   |
+| **`disableAutomaticViewportHeight`** | _boolean_ | If the content goes below the viewport do not increase the height so that it fits.                                                                        | `chrome.*`   |
```
oblador: Just noticed that the default value is wrong (also applies to some other values).

tjenkinson: Where is the default value?

oblador: It's at the end of the line `chrome.*`.

oblador:
Btw, I'm revamping the docs to use Docusaurus instead of Gitbook and noticed this is forked to the BBC org, is it OK to add the BBC logo on the website?

docs/configuration.md
tjenkinson: still not following

oblador: No worries, I can fix all of them in my docs revamp.

PR #174, opened 2019-10-02
Math.ceil the height value passed to setDeviceMetricsOverride

pawelk:
Some of my Loki tests are failing with 
`Failed with "Invalid parameters (height: integer value expected)" after 4 tries`
which is caused by non-integer `height` value being passed to `Emulation.setDeviceMetricsOverride()`.

This may happen when an element is sized in percentages.
Given `<div style="width:200px;padding-top:37%" />`, the result of `element.getBoundingClientRect().height` equals `166.11666870117188`. 
Similar construct is often used to create a fixe-ratio element that scales with the viewport.

If an element like that exists in the viewport, and the rendered area exceeds the pre-defined viewport dimensions, the screenshot can not be taken due to the forementioned error.

I have decided to apply `Math.ceil` only to the value passed to `setDeviceMetricsOverride` instead of rounding it in e.g. `getSelectorBoxSize` to not compound the rounding error; non-integer values do not cause harm in other places.


pawelk:
@oblador Thank you for such a quick approval. I have just added a test case in the React examples with a distilled version of a component that was failing for me due to this error.
BTW the failing CI build is due to `https://loki.js.org/favicon.png` returning a 404, unrelated to this PR :)

oblador:
@pawelk I know, I have a fix, just need it to pass CI before I make a PR :-)

oblador:
BTW, since you added a visual test you need to also commit the baselines.

oblador:
Fixed on master now, so just rebase

mulyoved:
Hit by this one, any chance to release with this fix?

oblador:
Released in [v0.18.1](https://github.com/oblador/loki/releases/tag/v0.18.1)

PR #179, opened 2019-11-04
Added flag --useRelativePath to use with --reactUri

jackall3n:
The work here gives someone the option to override the 'getAbsoluteUrl' method if it's not appropriate to their use case.


jackall3n:
@oblador any thoughts on this one?

oblador:
Hey, I'm trying to understand the use case :-) Why would you need a relative URL?

jackall3n:
Hey, so my particular use case is I want Loki to run inside a docker container on a CI server somewhere. As Loki runs in docker, it would be docker inside docker. 

To achieve this, CI docker (parent) and Loki docker (child) would need to share a volume. By default (and helpfully in most cases) Loki docker mounts the absolute path of the directory supplied. I thought it might be usefully if we could ask Loki in some cases not to manipulate the parameter. 

so the Loki docker run command:
docker run -v c:/foo/some-volume:/var

would become:
docker run -v some-volume:/var 

This change would not be solely for docker inside docker, but it would help it. 

Any thoughts?

oblador:
Would this use case be covered by this PR https://github.com/oblador/loki/pull/283?

oblador:
Closing for now, I think the copy argument should solve this. Please respond/reopen if you think otherwise.

PR #181, opened 2019-11-26
Support for remote docker hosts

Kristof-Mattei:
I change the `getNetworkHost` function because I felt it was incomplete.

It was catering for 2 use cases:

- loki runs natively, and Chrome in Docker
- loki runs in Docker, and Chrome is launched as a sibling

it was not catering for:

- loki runs natively, and Chrome in Docker on a different Docker host
- loki runs in Docker, and Chrome in Docker on a different Docker host

This pull request fixes that by first inspecting Docker's `context`, and if that one is not set to the local socket, we take its IP as the new container's host, on which we can make an ip:port connection. At this point it doesn't matter if we're dockerized.
If it is set to the local socket, proceed as in the past, i.e. see if we are dockerized or not.

This pull request was made to cater for docker builds on CircleCI, as CircleCI does not support siblings.

If you want to do Docker in Docker on CricleCI you need to do your 2nd level of Docker via `setup_remote_docker` which changes the context: https://circleci.com/docs/2.0/building-docker-images/

Notice that we can't just rely on `DOCKER_HOST` because the system may write to `~/.docker/config`. 

Reading from `docker context` is the most foolproof way.


Kristof-Mattei:
This won't work at all. We can't guarantee cross docker host communication as we don't control our own creation, so we can never call back to our storyboard.

PR #187, opened 2019-12-04
Pass storyId instead of kind+story

purelysloth:
This is a hacky solution to fix passing of `storyId` to `setCurrentStory`


src/targets/native/create-websocket-target.js
```diff
@@ -7,6 +7,29 @@ const { withTimeout, withRetries } = require('../../failure-handling');
 const MESSAGE_PREFIX = 'loki:';
 const NATIVE_ERROR_TYPE = `${MESSAGE_PREFIX}error`;
 
+const sanitize = (string) => {
```
purelysloth: taken from storybook

oblador:
Thanks for your PR! I just created a PR with latest RN & storybook to see what's broken, but it kinda works for me https://github.com/oblador/loki/pull/188

purelysloth:
Thanks for getting back quickly! Were you able to get it to work with multiple stories? Looking at the react native code that handles the `setCurrentStory` event, it looks like it only accepts `storyId` and not `kind` and `story`...

purelysloth:
https://github.com/storybookjs/storybook/blob/ba6b4c8ea0b4d1fc85baf030a8bfde0df9d39545/app/react-native/src/preview/index.tsx#L220

oblador:
I belive you, just trying to reproduce the issue so I can confirm that your PR works :-D My updated example uses RN 0.61.5 and SB 5.2.8, what versions are you on?

oblador:
I finally managed to reproduce and this does indeed fix it, thanks again!

purelysloth:
Thanks a lot for looking into it! :)

Issue #196, opened 2020-01-09
Loki fails to start when API calls from the components get an error


Geloloboy:
Is there a way for Loki to ignore API call errors when doing the tests?

oblador:
As of `0.19.0` there is now a [`fetchFailIgnore`](https://loki.js.org/configuration.html#fetchfailignore) config, thanks to @farisk <3!

PR #200, opened 2020-01-26
Allow warning for failed fetches rather than fail

farisk:
There may be valid reasons why a network request would fail for a
storybook, for example testing a failure case. Whilst ideally storybook shouldn't make requests you may want to use Loki before addressing that tech debt ( which is my use-case). To allow for this without
changing current behaviour, a "chromeWarnOnFetchFail" option has been
added. If it is enabled, rather than failing on fetch failure it prints
a warning message.

I've been using this change on prod, and someone else has posted the feature request > https://github.com/oblador/loki/issues/196

On a side note, having played around with printing stuff I found that things like "chromeRetries" are not in fact command line arguments but only work when put into the package.json.


farisk:
@oblador what do you think of this PR?

oblador:
Hi @farisk,

I recognize the use-case, but I'm not sure about the solution. What do you think about having a filter flag instead, so that you'd be excluding just the ones you expect to fail? Also the warn won't work with remote renderers such as the lambda stuff I'm working on. 

WDTY?

farisk:
Hey @oblador, being explicit with which ones you want to skip would be nice. How would the configuration look? I know that storiesFilter currently expects a regular expression, do you think it should be like that or be a list of story book names?

As for the warning message, is there something I can (or will be able to) hook into here to display it?

oblador:
I think it would make more sense for the filter to be URL based, and if it's more explicit on what urls to ignore I think we can skip the warnings. 

Btw, there's a few changes in your PR (the console imports) that look a bit off, is due to being out of sync with master?

oblador:
Thanks for the updates! Would you mind adding this option to the docs and [a story](https://github.com/oblador/loki/blob/master/examples/react/src/stories/index.js) to test the behavior?

farisk:
@oblador Hey. Thanks for the review. I have added a story and updated the documentation.

Geloloboy:
This is going to be a life saver. Hopefully we get this feature in as soon as possible :)

PR #210, opened 2020-02-20
AWS lambda renderer

oblador:
For very large test suite, running them on a single machine might be a bit slow. This PR introduces the concept of remote rendering, for now limited to AWS Lambdas to allow for a massive parallelisation (I run it at ~500 concurrent renders). There's a few params to tweak the execution/concurrency depending on your project.


packages/renderer-aws-lambda/src/index.spec.js
```diff
@@ -0,0 +1,9 @@
+const mod = require('./index');
```
techeverri: Just out of curiosity 😬 

What does `mod` mean? 🤔

packages/target-chrome-core/src/create-chrome-target.js
```diff
@@ -176,7 +178,7 @@ function createChromeTarget(
       }
     };
 
-    client.captureScreenshot = withRetries(options.chromeRetries)(
+    client.captureScreenshot = withRetries(options.chromeRetries, 500)(
```
techeverri: We could give `500`  a nice name and move it to the top of the module

What do you think? 🤔

packages/target-native-android-emulator/src/create-android-emulator-target.js
```diff
@@ -1,9 +1,17 @@
...
+const tempy = require('tempy');
 const osnap = require('osnap/src/android');
 const { createWebsocketTarget } = require('@loki/target-native-core');
 
-const saveScreenshotToFile = filename => osnap.saveToFile({ filename });
+const captureScreenshot = async () => {
+  const filename = tempy.file({ extension: 'png' });
+  await osnap.saveToFile({ filename });
+  const screenshot = await fs.readFile(filename);
+  await fs.unlink(filename);
```
techeverri: What if `fs.readFile` or  `fs.unlink` fail?

Should this be wrapped in a `try-catch` block to handle (e.g. log, report, etc) the error?

packages/renderer-aws-lambda/src/index.spec.js
oblador: `module`, but it's a reserved word, so you can't/shouldn't have a variable with that name.

packages/target-native-android-emulator/src/create-android-emulator-target.js
oblador: The try-catch happens on the runner level, if the screenshot fails the story will fail which is correct behavior IMHO. You can't really continue with the comparison after this.

packages/target-chrome-core/src/create-chrome-target.js
oblador: Fixed!

packages/target-native-android-emulator/src/create-android-emulator-target.js
techeverri: If there is already a top-level `try-catch` block, it makes sense :+1:

PR #211, opened 2020-02-20
Fix looks-same config params reading

raulmt:
Hi!

It seems #127 had a bug when it was merged. 

The docs say the config parameter in `package.json` should be `looks-same` (https://github.com/oblador/loki/blame/master/docs/configuration.md#L48), but currently that's not what gets read.

[Here](https://github.com/oblador/loki/blame/master/docs/configuration.md#L48) it's reading `looksSame` instead of `looks-same`, and storing it as a key `looksSame` also in the parsed configuration.

Later, when creating the diffing engine [here](https://github.com/oblador/loki/blob/master/packages/runner/src/commands/test/test-story.js#L49-L52), the it tries to get a config key with the same name as the `diffingEngine` given, which is never `looksSame`, but instead is either `gm` or `looks-same`. As a result, the config for `looks-same` never gets passed through.

This PR just changes the config parsing to read it the same way as it got documented, and stores it in the parsed config the same way too, since it will get read that way.

There are other options to fix this, but this one seemed to be the less invasive one.

### DIFF
--- a/packages/runner/src/commands/test/parse-options.js
+++ b/packages/runner/src/commands/test/parse-options.js
@@ -40,7 +40,7 @@ function parseOptions(args, config) {
     diffingEngine:
       $('diffingEngine') || (dependencyAvailable('gm') ? 'gm' : 'looks-same'),
     fetchFailIgnore: $('fetchFailIgnore'),
-    looksSame: $('looksSame'),
+    'looks-same': $('looks-same'),
     gm: $('gm'),
     verboseRenderer: $('verboseRenderer'),
     requireReference: $('requireReference') || ciInfo.isCI,


oblador:
Thanks!

raulmt:
@oblador thanks for looking into this so quickly!

Any chance you can do a release? Not only because of this, but on the monorepo migration it looks like you upgraded looks-same also. That brings `ignoreCaret` option which I need (I have a couple of stories with a form that has autoFocus on one field, and loki randomly fails depending on whether the screenshot moment occurred with respect to the caret blinking :( ). Actually, I noticed this bug because I was trying to make that option work 😂 and when I fixed this, it still didn't work because of the older version 😞 

For the moment I'm forcing the looks-same version through a yarn `resolution`, and it seems to be working for now…

oblador:
It's released in [0.19.0](https://github.com/oblador/loki/releases/tag/v0.19.0) along with a bunch of other things!

raulmt:
@oblador Thanks a lot!

PR #218, opened 2020-03-06
Wait between retries when launching stories tab.

Kristof-Mattei:
This fixes a common failure when doing `loki` on CircleCI. 

Over 100 runs:
* 100 times would fail the first time (caught by the `withRetries(2`)
* 90 times would work the second time
* 10 times it would fail the second time

This tries to fix 10% of the cases. 

Over the last 2 weeks we have been testing this patch (> 200 builds) and have had ZERO failures!

### DIFF
--- a/packages/target-chrome-core/src/create-chrome-target.js
+++ b/packages/target-chrome-core/src/create-chrome-target.js
@@ -17,6 +17,7 @@ const {
 } = require('@loki/core');
 const presets = require('./presets.json');
 
+const RETRY_LOADING_STORIES_TIMEOUT = 10000;
 const LOADING_STORIES_TIMEOUT = 60000;
 const CAPTURING_SCREENSHOT_TIMEOUT = 30000;
 const CAPTURING_SCREENSHOT_RETRY_BACKOFF = 500;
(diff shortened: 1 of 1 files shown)


Kristof-Mattei:
Any comment on this?

packages/target-chrome-core/src/create-chrome-target.js
```diff
@@ -17,6 +17,7 @@ const {
 } = require('@loki/core');
 const presets = require('./presets.json');
 
+const RETRY_LOADING_STORIES_TIMEOUT = 10000;
```
oblador: This value seems a bit high, how did you come up with it?

oblador: I'll merge it as is now and we can follow up with tweaks if needed.

Kristof-Mattei: I would’ve taken a lower value, but I would only do that if we did something like an exponential back off, start slow and iterate maybe 10 times, doubling the wait every time.

oblador:

PR #219, opened 2020-03-11
don't clear error logs.

fleck:
When there's an error starting Chrome I get:
`FAIL chrome.docker/START: Cannot read property 'length' of null`
instead of the actual error message.

Not sure if there was a reason to set errorMessages to null, but this fixes the problem.


oblador:
I think the reasoning I had was to not hog memory for unused logs, but it shouldn't be that much anyways. An alternative would be to set it to `[]` or have a null check.

oblador:

PR #220, opened 2020-03-12
Add storiesOf parameters (third argument) support

raulmt:
According to [the docs](https://storybook.js.org/docs/formats/storiesof-api/#decorators-and-parameters), storybook supports a third parameter when calling `add`, to give story-level parameters (which can be options, config or parameters to send to decorators, for example).

Without this PR, it's not possible to pass parameters to a decorator when using lokiAsync. This PR enables that, and shouldn't cause a problem to eventually older storybook versiones that don't support that due to the nature of non-declared parameters in JavaScript ;)

Let me know what you think please


oblador:

PR #223, opened 2020-03-25
Disable pointer events

oblador:
As reported by @lindesvard, occasionally elements positioned at the 0x0 position display hover styling. I'm not really able to reproduce myself but this is a proposed fix that also fixes a chrome bug with negatively positioned elements.


packages/target-chrome-core/src/create-chrome-target.js
```diff
@@ -161,6 +162,7 @@ function createChromeTarget(
         debug('Disabling animations');
         await evaluateOnNewDocument(`(${disableAnimations})(window);`);
       }
+      await evaluateOnNewDocument(`(${disablePointerEvents})(window);`);
```
techeverri: Is there a use case where people would like to "enable pointer-events" just like the `chromeEnableAnimations` option above?

If so, there should be a way (e.g. flag from `package.json` or command line) to enable/disable it 👍

oblador: I honestly don't think so. Disabling animations is much more invasive and could potentially break apps, this shouldn't.

Issue #221, opened 2020-03-13
Loki binary not being registered


aarongarciah:
Since `0.19.0`, Loki's binary fails to be registered. This is the output when installing Loki via `yarn global` using every version from `0.18.1` until `0.21.1`, see how yarn warns about the lack of binary being installed:

### Loki 0.18.1

<img width="378" alt="Screenshot 2020-03-13 at 08 10 31" src="https://user-images.githubusercontent.com/7225802/76599701-d1f5a500-6505-11ea-9831-beabfac7d0be.png">

### Loki 0.19.0

<img width="316" alt="Screenshot 2020-03-13 at 08 14 04" src="https://user-images.githubusercontent.com/7225802/76599704-d326d200-6505-11ea-8ac8-f24e8053eb7c.png">

### Loki 0.19.1

<img width="314" alt="Screenshot 2020-03-13 at 08 22 59" src="https://user-images.githubusercontent.com/7225802/76599705-d326d200-6505-11ea-8c16-72db67ddeffb.png">

### Loki 0.20.0

<img src="https://user-images.githubusercontent.com/7225802/76599696-d02be180-6505-11ea-827f-93b43d47f579.png" width="310" />

### Loki 0.20.1

<img width="312" alt="Screenshot 2020-03-13 at 08 11 00" src="https://user-images.githubusercontent.com/7225802/76599703-d28e3b80-6505-11ea-828d-70ef25076abf.png">

This is problematic when installing Loki in a Docker image and you want to use the binary command and it doesn't exist. You need to locate the binary path by hand to execute it.

Thanks for your hard work @oblador. Keep up the great work!

oblador:
Fixed in [0.20.2](https://github.com/oblador/loki/releases/tag/v0.20.2].

aarongarciah:
Thank you @oblador!

PR #224, opened 2020-03-25
Add loki command alias to entry package

oblador:
With the monorepo restructuring the binary was moved to a subpackage, which seems to cause warnings/errors when installing globally as outlined in #221. 

This PR adds it back to the main `loki` package as an alias to the `runner` package.

### DIFF
--- a/packages/loki/bin/loki
+++ b/packages/loki/bin/loki
@@ -1,0 +1,3 @@
+#!/usr/bin/env node
+
+require('@loki/runner/src/cli')();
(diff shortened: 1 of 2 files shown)


techeverri:
Is there a way to verify this without publishing a new version?

oblador:
I ran the script itself locally, but reproducing the `yarn global add` would first require us to publish. My main concern is that we now have a naming conflict as we have to packages exposing a `loki` command. However it was like this before by accident too, so it should work fine.

PR #226, opened 2020-03-30
improve getSelectorBoxSize to find fixed & absolute elements

lindesvard:
If we had a story with absolute position and the chromeSelector element had height 0 `getSelectorBoxSize` wouldn't find the positioned element.

This improves this. Have added a test to show it as well.


oblador:
Pushed a few new possible scenarios to your story that we should support.

oblador:
Care to have a look at this one @techeverri?

techeverri:
Will have a look at it tomorrow 👍

packages/browser/src/get-selector-box-size.js
```diff
@@ -1,26 +1,153 @@
...
+  function isElementHiddenByOverflow(
+    element,
+    { hasParentFixedPosition, hasParentOverflowHidden }
+  ) {
+    const checkOutOfBounds = () => {
+      try {
+        const elementRect = element.getBoundingClientRect();
+        const containerRect = hasParentOverflowHidden.getBoundingClientRect();
+        const top = elementRect.top < containerRect.top;
+        const bottom = elementRect.bottom > containerRect.bottom;
```
techeverri, line 30: Maybe I don't understand
But shouldn't this be "less than" to be outside the container?

```suggestion
        const bottom = elementRect.bottom < containerRect.bottom;
```

packages/browser/src/get-selector-box-size.js
```diff
@@ -1,26 +1,153 @@
...
+  function hasFixedPosition(element) {
     const style = window.getComputedStyle(element);
+    return style.position === 'fixed';
+  }
+
+  function isElementHiddenByOverflow(
+    element,
+    { hasParentFixedPosition, hasParentOverflowHidden }
+  ) {
+    const checkOutOfBounds = () => {
```
techeverri, line 25:
```suggestion
    const isElementOutOfBounds = () => {
```

packages/browser/src/get-selector-box-size.js
lindesvard: This is correct. If the child's bottom part is greater than the container's bottom we don't want that element.

packages/browser/src/get-selector-box-size.js
lindesvard: fixed

packages/browser/src/get-selector-box-size.js
techeverri: I see 👍

techeverri: Shouldn't all the comparisons be "greater than" then? 🤔
I guess I still don't get it 🙈 

Could you please elaborate more on how this works?

lindesvard: Hope this helps to explain: https://imgur.com/a/AB84p4m

techeverri: I finally get it 😅 

Got confused because of the similarities between [element positioning](https://developer.mozilla.org/en-US/docs/Web/CSS/position) and the returned value of [`getBoundingClientRect`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect). Both use the same `top, right, bottom, left` names but with different meanings.

Positioning | getBoundingClientRect
------------ | -------------
unnamed | element-box-diagram

Issue #103, opened 2018-08-22
Problems caused by skip() & lokiSkip(): they break other storybook plugins


jacksbox:
`lokiSkip()` and `asyc` are pretty handy functions, but they break other storybook plugins:
e.g. the storyshots plugin, which complains, that `lokiSkip` isn't a function.
this happens because loki just registers this functions when it is configuring storybook.
All other plugins, don't use the loki `config-storybook` and will therefor break.

Please find another possibility to add this functions (just register a normal storybook plugin?) or keep the cli option to exclude stories.

erzhtor:
I'm not sure is it related, but I also faced problem while passing `@storybook/addon-viewport` parameters to `.lokiSkip`.

Example:
```jsx
storiesOf('Card', module)
    .add('default', () => (
        <Card />
    ))
    .lokiSkip('iphone6', () => (
        <Card />
    ), { viewport: { defaultViewport: 'iphone6' }})
```

Parameters are not recognized and both use the same viewport. But using `.add` instead of `.lokiSkip` works as expected and sets viewport according to parameters.

erzhtor:
Any update there?

Tobi-mmt:
I have the same problem. Are there any updates?

## Setup
Node v12.16.2
Loki v0.22.1
Storybook v5.3.13
Storybook addons: `docs`, `viewport`

## Example
```js
storiesOf('Card', module)
    .addParameters({ component: Card })
    .add('default', () => (
        <Card />
    ))
    .lokiSkip('default2', () => (
        <Card />
    ))
```

## What do I expect?
If I start storybook there should be two stories. If I go to the Docs tab in storybook there should be the documentation of all two stories, including the props list of the component on top and the code of each story visible next to each rendered story.
Cod button example in right bottom corner: 
![Bildschirmfoto 2020-07-01 um 09 07 05](https://user-images.githubusercontent.com/14888529/86213632-4d916580-bb7a-11ea-8873-23f456b89be2.png)


## What is really happening?
I can not see the code from the `default2` story, but I can from the `default` story. 
Code button: 
![Bildschirmfoto 2020-07-01 um 09 07 00](https://user-images.githubusercontent.com/14888529/86213660-584bfa80-bb7a-11ea-8915-ef3b9f0328c8.png)


## Suggestions
This looks like `lokiSkip` prevents some functionality of the add-ons. 
It should be possible to pass the whole `add` function through the `lokiSkip` to prevent

Issue #198, opened 2020-01-18
Incompatibility of story names on Storybook 5.3


tassoevan:
This is probably related to #190. From https://github.com/storybookjs/storybook/blob/master/MIGRATION.md#deprecate-displayname-parameter:

> In 5.3, the CSF loader decouples the story's name/id, which means that displayName is no longer necessary. Unfortunately, this is a breaking change for any code that uses the story name field. Storyshots relies on story name, and the appropriate migration is to simply update your snapshots. Apologies for the inconvenience!

In real life, it means no story names containing JavaScript keywords (e.g. `default`, `with`) and camelCase/PascalCase identifiers (e.g. `onHover`) because that story ID diverges from the name.

Issue #190, opened 2019-12-06
CSF stories using the "name" parameter always fail


JohnAlbin:
I have a super simple component which renders perfectly in Storybook 5.2. It uses the CSF format and the "name" parameter so that I can use a different name from the named export name.

```JS
import React from 'react';
import Error from './index';

export default {
  component: Error,
  title: 'Pages/Error',
}

export const Error404 = () => (
  <Error
    statusCode={404}
  />
);
Error404.story = {
  name: '404 Error',
};
```

In CSF, the default name for a story is its exported variable name with spaces inserted before camelcase letters or numbers. In this case it's `Error 404`, but the name I really want is "404 Error" since that's its common terminology. I can't use `export const 404Error` because that is invalid JavaScript. Fortunately, CSF allows you to override the default export name with the `[export var].story.name` parameter.

This option works great inside Storybook. Unfortunately, it causes Loki to fail to make a screenshot saying:
```
      ❯ Pages/Error
        ✖ 404 Error
          → Failed with "Unable to get position of selector "#root > *". Review the `chromeSelector` option and make sure your story doesn't crash.
…
```

JohnAlbin:
Update: I just noticed I have another story with `TODO.story = { name: "@TODO"}` and that renders fine with Loki.

So perhaps its the use of a space in the name that cause Loki to fail?

PR #228, opened 2020-04-02
Support skip via story parameters

### LINKED ISSUE #176: No way to skip a story using CSF (Storybook 5.2)
When writing stories using the [CSF format](https://storybook.js.org/docs/formats/component-story-format/), there is no way to mark a single story to be skipped. Ideally, a parameter would be neat:

```
export const componentVariation = () => <MyComponent prop1={val1} />;
componentVariation.story = {
  parameters: { skipLoki: true },
};
```

oblador:
This PR moves the configuration of skipping individual stories to the story `parameter` field, deprecating the `lokiSkip` method. Additionally it drops the deprecated use of `kind+story` combination and uses the `id` introduced in SB 5. 

Fixes #176 #190 #198

**This is a breaking change**, dropping support for storybook prior to version 5.0 which was released about one year ago. It is probably possible to keep backwards support, but it would be a bit messy and would have to rely on heuristics.


examples/react/src/stories/index.stories.js
```diff
@@ -17,9 +17,11 @@ import FetchComponent from '../FetchComponent';
 import ZeroHeightWithPadding from '../ZeroHeightWithPadding';
 import Hover from '../Hover';
 
-storiesOf('Welcome', module).lokiSkip('to Storybook', () => (
-  <Welcome showApp={linkTo('Button')} />
-));
+storiesOf('Welcome', module)
+  .lokiSkip('to Storybook', () => <Welcome showApp={linkTo('Button')} />)
```
techeverri, line 21: So `lokiSkip` will continue to exist and behave like `add`?

packages/integration-core/src/decorate-storybook.js
```diff
@@ -8,23 +8,23 @@ let warnedAsyncDeprecation = false;
...
-  function wrapWithSkipStory(add, kind, isDeprecatedCall) {
-    return function skipStory(story, storyFn, parameters) {
-      if (isDeprecatedCall && !warnedSkipDeprecation) {
+  function wrapWithSkipStory(add) {
+    return function skipStory(story, storyFn, parameters = {}) {
+      if (!warnedSkipDeprecation) {
         warnedSkipDeprecation = true;
         console.warn(
-          '[DEPRECATED] `.add.skip(...)` is deprecated. Please use `.lokiSkip(...)` instead.'
+          '[DEPRECATED] `.add.skip(...)` and `.lokiSkip(...)` are deprecated. Please pass `{ loki: { skip: true } }` in the third parameter instead.'
```
techeverri, line 17: Nice warning! 💯

examples/react/src/stories/index.stories.js
oblador: Yeah, i'll add warnings for now and remove it some time in the future.

examples/react/src/stories/csf.stories.js
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+
+import { storiesOf } from '@storybook/react';
+import { linkTo } from '@storybook/addon-links';
```
techeverri, line 4: Does the example need these 2 imports? 🤔

oblador: Fixed!

Tobi-mmt:
@oblador this will probably also close #103 – if we use the parameters instead of `skip()` – since the stories are filtered by there parameters without an blocking, extra function.

I am excited for this update 🥳

johannespfeiffer:
Here too 😄 
Any eta for this update?

PR #241, opened 2020-06-03
Make blinking input carets transparent to avoid flakiness

oblador:
Will still preserve focused styling if one wants to test that.


techeverri:
Does it work in all platforms (iOS, Android and web)?

oblador:
This is browser only unfortunately :-(

donaldpipowitch:
Oh that is wonderful. Can't wait for a new release 😻

PR #246, opened 2020-06-16
Add isLokiRunning()

oblador:
This PR adds a new package `@loki/is-loki-running` with a function that returns `true` during a test run, otherwise `false`. This is useful to disable things causing flakiness, such as JavaScript animations that aren't automatically disabled by loki. 

Improves #110 and fixes #111.


packages/is-loki-running/src/index.native.js
```diff
@@ -0,0 +1,4 @@
+module.exports = function isLokiRunning(win = global) {
+  // eslint-disable-next-line no-underscore-dangle
```
techeverri, line 2: This comment doesn't seem to apply to the line below

packages/is-loki-running/src/index.web.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-env browser */
+
+module.exports = function isLokiRunning(win = window) {
+  // eslint-disable-next-line no-underscore-dangle
```
techeverri, line 4: This comment doesn't seem to apply to the line below

packages/is-loki-running/src/index.native.js
oblador: Fixed in https://github.com/oblador/loki/commit/472067f089525ef163acc31dc00a0311bca9c4d5

PR #247, opened 2020-06-22
Fix loki global check for React Native

levibuzolic:
It appears that the `global.loki` check added in #246 is incorrect and should instead be trying to create the `global.loki` only if it's not already defined.

This is currently causing a error preventing React Native tests from running at all as the line immediately after the if statement will try and assign `global.loki.isRunning = true;` while `global.loki` is undefined.


oblador:
Thanks!

oblador:
Released in [0.22.1](https://github.com/oblador/loki/releases/tag/v0.22.1).

PR #248, opened 2020-06-26
Add Rocket.Chat to Users of Loki

MartinSchoeler:
Hello! We are currently using Loki on our Storybook ([Rocket.Chat's Fuselage](https://github.com/RocketChat/Rocket.Chat.Fuselage)) So we thought it would be nice to be on the list of users. Thanks!


oblador:
Really cool, thanks!

PR #257, opened 2020-07-24
Remove version in node requirement

### LINKED ISSUE #230: Upgrade minimum required Node.js version to an active LTS or current
### Context

According to the [docs](https://github.com/oblador/loki#prerequsites) and the [`package.json`](https://github.com/oblador/loki/blob/caa4742573e8f98cb79d08428809eddff1e4cbdb/package.json#L16), the minimum required version of Node.js for Loki  is 7.6 or higher

### Issue

Node.js 7.x was released [2016-10-25](https://nodejs.org/en/download/releases/#:~:text=Node.js%207.0.0) and end-of-life (last commit) happened [2017-07-11](https://github.com/nodejs/node/commits/v7.x)

Considering that Node.js 7.x is officially unsupported and that there is no point on supporting deprecated versions we could upgrade the minimum required version to an active TLS 
e.g. Node.js 10.x end-of-life will be 2021-04-30 (more info table and link below)

image
🔗 https://nodejs.org/en/about/releases/

oblador:
Fixes #230

### DIFF
--- a/docs/getting-started.md
+++ b/docs/getting-started.md
@@ -5,7 +5,7 @@ title: Getting started
 
 ## Prerequsites
 
-- Node 7.6 or higher
+- [Node](https://nodejs.org/)
 
 ### Optional dependencies
 
--- a/packages/loki/README.md
+++ b/packages/loki/README.md
@@ -22,7 +22,7 @@ Loki aims to have easy setup, no to low maintenance cost, reproducible tests ind
 
 ## Prerequsites
 
-- Node 7.6 or higher
+- [Node](https://nodejs.org/)
 
 ### Optional dependencies
 


techeverri:
There is at least one more reference in code to the minimum required Node.js version

https://github.com/oblador/loki/blob/8f5f5fe56b45a762beac1ad3bf36b3cf5f93c9a0/package.json#L17

Some other references to Node.js version are in the following files:
- https://github.com/oblador/loki/blob/master/circle.yml
- https://github.com/oblador/loki/blob/master/.travis.yml

oblador:
Circle isn't used right now, but for travis it could make sense to have a build matrix perhaps. 

Root package isn't deployed so doesn't have much effect, but also isn't necessarily incorrect?

PR #273, opened 2020-10-06
Fix graphicsMagick comparing images with different dimensions. Fixes #44

### LINKED ISSUE #44: visual tests pass when a component's size changes
Steps to reproduce: 

1. create a 150x150 blue square component.
2. run loki update
3. change the size to 300x300
4. run loki test
Expected: test should fail
Actual: test is passing

Crazymax11:
## PRE

Here is no contributing guide, so I don't know how to format Pull Request, which steps I need to do to check my changes and I don't know how to test this changes properly.

## Problem

GraphicsMagick's compare method checks only pixel diff in first image dimensions. This means that if second image is first image concated with another image, GM will decide that the images are equal.

In practice it means, that than I, as programmer, add some content to existing component/story, GM will not report any diff and Loki.js tests will pass. 

## Resolution

Check images dimensions manually. That method I use in current job project and it works.

### DIFF
--- a/packages/diff-graphics-magick/src/create-graphics-magick-differ.js
+++ b/packages/diff-graphics-magick/src/create-graphics-magick-differ.js
@@ -1,9 +1,21 @@
 const fs = require('fs-extra');
 const gm = require('gm');
+const imageSize = require('image-size');
 
 function createGraphicsMagickDiffer(config) {
   return function getImageDiff(path1, path2, diffPath, tolerance) {
     const instanceConfig = { tolerance: tolerance / 100, ...config };
+
+    const path1Dimensions = imageSize(path1);
+    const path2Dimensions = imageSize(path2);
+
+    if (
+      path1Dimensions.height !== path2Dimensions.height ||
+      path1Dimensions.width !== path2Dimensions.width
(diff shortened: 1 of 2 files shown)


packages/diff-graphics-magick/src/create-graphics-magick-differ.js
```diff
@@ -1,9 +1,21 @@
 const fs = require('fs-extra');
 const gm = require('gm');
+const imageSize = require('image-size');
 
 function createGraphicsMagickDiffer(config) {
   return function getImageDiff(path1, path2, diffPath, tolerance) {
     const instanceConfig = { tolerance: tolerance / 100, ...config };
+
+    const path1Dimensions = imageSize(path1);
```
oblador, line 9: Can't we use [`gm.size`](https://aheckmann.github.io/gm/docs.html#getters) instead so we can lose the added dependency?

Crazymax11: replaced image-size with gm.size

oblador (approved):

oblador:
Thank you!

Issue #229, opened 2020-04-02
ENOENT error encountered while using fileNameFormatter feature


Geloloboy:
DESCRIPTION:
The following error is encountered when running into an image with valid difference from baseline:
![Loki Filename Error](https://user-images.githubusercontent.com/25830532/78304245-051dd800-758a-11ea-8a56-70cd7400fe7c.png)

STEPS TO REPLICATE:
Set the following attribute in loki.config.js:
`fileNameFormatter: ({ configurationName, kind, story }) = `${configurationName}/${kind}_${story}`.toLowerCase()`

EXPECTED RESULT:
An image is different error is encountered
Difference is shown on the difference folder

ACTUAL RESULT:
ENOENT error is encountered
No image gets saved on the difference folder

NOTE:
Haven't tested on a flat file structure since our storybook instance have folders set up in them

jalooc:
> Haven't tested on a flat file structure

It works for flat structure, so there's definitely a bug related to missing directories when creating diff. +1 cause this currently makes using `fileNameFormatter` impossible for most usages.

jalooc:
After a short research I think I found it:

Loki uses [`looksSame.createDiff`](https://github.com/oblador/loki/blob/b06e7512dcc2314b977571013fa97b59c1b898a8/packages/diff-looks-same/src/create-looks-same-differ.js#L29) passing the `diff` option to `looks-same` in the form of the path to the file. `looks-same` in turn, after some processing [saves](https://github.com/gemini-testing/looks-same/blob/5406adfb1e9a7b8e61c3224addc536320cbb9897/index.js#L221) the file using [`fs.createWriteStream()`](https://github.com/gemini-testing/looks-same/blob/5406adfb1e9a7b8e61c3224addc536320cbb9897/lib/png/png.js#L49), passing that path as a parameter. Now, `fs.createWriteStream()` fails when the directory the path is pointing to doesn't exist - and that's the problem.

The reason why saving _reference_ and _current_ images works in Loki is because Loki [saves them itself using `fsExtra.outputFile()`](https://github.com/oblador/loki/blob/b06e7512dcc2314b977571013fa97b59c1b898a8/packages/runner/src/commands/test/compare-screenshot.js#L33) which creates any missing directories.

[...]

PR #276, opened 2020-10-09
Allow slashes in diff paths with looks-same

kurttomlinson:
This PR is a partial fix for https://github.com/oblador/loki/issues/229. It creates the output diff folder immediately before asking looks-same to create a diff file. Because it only creates the folder when a diff file is going to be created, it does not create any empty folders.

This patch only applies to the looks-same differ. The Graphics Magic differ creates diff files even when there are no differences, so it needs to be fixed in a different way.

### DIFF
--- a/packages/diff-looks-same/src/create-looks-same-differ.js
+++ b/packages/diff-looks-same/src/create-looks-same-differ.js
@@ -20,12 +20,13 @@ function createLooksSameDiffer(config) {
         return reject(new Error('Current image is empty'));
       }
 
-      return looksSame(reference, current, instanceConfig, (err, isSame) => {
+      return looksSame(reference, current, instanceConfig, async (err, isSame) => {
         if (err) {
           reject(err);
         } else if (isSame) {
           resolve(isSame);
         } else {
+          await fs.ensureFile(diffPath);
           looksSame.createDiff(
             {
               ...instanceConfig,
--- a/packages/diff-looks-same/src/create-looks-same-differ.js
+++ b/packages/diff-looks-same/src/create-looks-same-differ.js
@@ -20,13 +20,13 @@ function createLooksSameDiffer(config) {
         return reject(new Error('Current image is empty'));
       }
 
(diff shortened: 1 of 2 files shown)


techeverri (commented):
> I think a better approach would be to write the diff to a temp dir and then move it if there is a diff. Otherwise we'd end up with a bunch of empty directories.

~Based on the comment above https://github.com/oblador/loki/issues/229#issuecomment-663382861 by @oblador I would say this is not the desired behavior~

techeverri:
I would recommend you to use the workaround https://github.com/oblador/loki/issues/229#issuecomment-699680908 suggested by @SebDuf using `patch-package` in #229

kurttomlinson:
@techeverri I don’t think that comment applies in this case. There is an `if (isSame)` case above that handles when the images are the same. This newly added line of code is only executed when we already know there is a difference and a diff file needs to be created. This PR will not create any unnecessary folders.

techeverri:
You're right @kurttomlinson 👍

techeverri:
It would be very useful to have some kind of automated test for this fix 🤔 

I'm thinking of an example project configured to have custom `fileNameFormatter` that uses `/` slashes (subfolders)

techeverri:
BTW because of the way GitHub Actions work builds won't run when they come from forked repos (this pull request for example)
Recently I learned we shouldn't merge without running the changes through the CI pipeline

kurttomlinson:
@techeverri I've added a spec to cover this change.

kurttomlinson:
@techeverri Sorry to be a bother, but is there something I could do to help you review and merge this PR? 🙏

oblador:
@kurttomlinson I've fixed so that tests run properly now 👍

oblador (approved):

kurttomlinson:
Thanks for the fix and the merge @oblador ! 🎉

oblador:
Thanks for the PR! I was having similar thoughts actually :-D

PR #283, opened 2020-11-11
allow user to use docker cp instead of volume mount, fix #9

### LINKED ISSUE #9: Running loki in CI setup
Hi,

What is the best way to run loki in CI like CircleCI or Codeship? 

Can it run across different simulators/emulators?

Thanks,
Leena

missing1984:
## Context

This should fix one of the long live issues. https://github.com/oblador/loki/issues/9

Docker volume mount does not always work especially in a CI environment that often implemented as docker in docker. 

## Solution

Introduce a new flag `chromeDockerUseCopy ` that switches to use docker cp which is guaranteed to work in CI. Users can opt-in to this new behavior.

I also fixed a couple of dependencies issues in this MR.


missing1984:
@techeverri mind take a look at this when you're free :) ?

techeverri:
Hey @missing1984 
I don't have much experience with Docker in Docker stuff. 
So I will leave this one for @oblador 👍

packages/integration-react-native/package.json
```diff
@@ -22,6 +22,10 @@
   "dependencies": {
     "@loki/integration-core": "^0.25.0"
   },
+  "peerDependencies": {
```
oblador, line 25: The reason i don't have these is because I don't want to show warnings for people using loki with regular react or with vue. You might argue that it shouldn't be in the main `loki` package, but it's there for legacy reasons. I would prefer if you removed these `peerDependency` declarations as I don't think anybody would use loki for react-native or react without also having storybook.

oblador: In other words: your approach, is technically more correct, but I doubt it would help any users that forgot adding storybook as a dependency, but instead it would confuse a lot of users that have everything set up correctly because `loki` depends on these.

packages/target-chrome-docker/src/create-chrome-docker-target.js
```diff
@@ -67,12 +71,14 @@ function createChromeDockerTarget({
...
-  } else if (dockerUrl.indexOf('file:') === 0) {
-    const staticPath = dockerUrl.substr('file:'.length);
-    const staticMountPath = '/var/loki';
-    runArgs.push('-v');
-    runArgs.push(`${staticPath}:${staticMountPath}`);
+  } else if (isLocalFile) {
+    localPath = dockerUrl.substr('file:'.length);
     dockerUrl = `file://${staticMountPath}`;
+    if (!chromeDockerUseCopy) {
+      // setup volumn mount if we're not using copy
```
oblador, line 78:
```suggestion
      // setup volume mount if we're not using copy
```

packages/target-chrome-core/package.json
```diff
@@ -21,7 +21,9 @@
   "main": "src/index.js",
   "dependencies": {
     "@loki/browser": "^0.25.1",
-    "@loki/core": "^0.25.0"
+    "@loki/core": "^0.25.0",
+    "@loki/integration-core": "^0.25.0",
+    "debug": "^4.2.0"
```
oblador, line 26:
```suggestion
    "debug": "^4.1.1"
```
This way we only have one version of this dependency defined in the workspace. If we're gonna bump, I'd argue we bump all packages.

packages/integration-react-native/package.json
missing1984: I think a library should always declare their dependencies based on the package.json contract rather than the use cases as the latter is highly subecjtive. 

Would declare those peerDependnecies  as optional by setting `optional:true` in `peerDependenciesMeta` an option for you?

npm v6+, yarn and pnpm are all recognized this flag, https://classic.yarnpkg.com/en/docs/package-json/#toc-peerdependenciesmeta

oblador: Ah, yes that would work!

packages/target-chrome-core/package.json
```diff
@@ -20,8 +20,10 @@
   ],
   "main": "src/index.js",
   "dependencies": {
+    "@loki/integration-core": "^0.25.0",
```
oblador, line 23:
```suggestion
    "@loki/integration-core": "^0.26.0",
```

packages/integration-react-native/package.json
missing1984: Fixed!

missing1984: @oblador not mean to push you for this :) just wondering if there are other concerns of this MR that I should address?

oblador: You missed two other packages, but fixed those.

oblador (approved):

PR #285, opened 2020-11-24
Run tests on PRs

oblador:
I'm somewhat new to the github actions thingie, and I made a boo-boo.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -23,13 +23,8 @@ jobs:
         run: yarn --frozen-lockfile --non-interactive --silent --ignore-scripts
       - name: Prepare docker
         run: docker pull yukinying/chrome-headless-browser:85.0.4181.8
-      - name: Run Jest
-        uses: tangro/actions-test@1.1.6
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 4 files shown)


techeverri:
As far as I know GitHub Actions run only for contributors

techeverri:
Or at least that's what I noticed from the last pull request submitted

oblador:
I've now changed the settings so that we can't merge anything without tests passing. Will rebase other PRs.

.github/workflows/tests.yml
```diff
@@ -1,6 +1,12 @@
 name: Tests
 
-on: [push]
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
techeverri, line 9: After this change the test don't run on push to branches without pull request

Is there a way to have both for the pull request but also for branches while developing

oblador: I couldn't find one easily. Problem is, if you don't have this it will run tests twice for branches in the same repo. Quick hack is to remove this line in your branch and add it back before making the PR.

techeverri: Nice trick 👍 

_I'll try removing the code while pushing to a branch and remove it once the pull request is open_

PR #291, opened 2020-12-24
add storybook 6 in peerDependencies list

missing1984:
Tested with storybook 6 and it's working great.

Adding storybook 6 in our peerDependencies list.

also `loki` claims `@loki/integration-react` as its dependencies, we would have to decalre optional peerDepndnecies for `loki` as well.


oblador (approved):
Looks good, thx!

PR #292, opened 2021-01-03
Add meta data for Storybook Addon Catalog

oblador:
Per instructions from this page: https://storybook.js.org/docs/react/addons/addon-catalog

### DIFF
--- a/packages/loki/package.json
+++ b/packages/loki/package.json
@@ -1,14 +1,16 @@
 {
   "name": "loki",
   "version": "0.27.0",
-  "description": "Visual Regression Testing for react-storybook",
+  "description": "Visual Regression Testing for Storybook",
   "bin": {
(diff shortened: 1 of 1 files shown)


techeverri (approved):

packages/loki/package.json
```diff
@@ -18,6 +20,10 @@
     "android",
     "web"
   ],
+  "storybook": {
+    "displayName": "Loki",
+    "icon": "https://loki.js.org/img/favicon.svg"
+  },
```
techeverri, line 26: What about setting `unsupportedFrameworks` and `supportedFrameworks` accordingly?

oblador: I had it from the beginning, but was thinking that it should already support all frameworks out of the box. Maybe it makes sense to add it anyways for things we've tried and tested with.

PR #293, opened 2021-01-03
Bump react-native example to 0.63/storybook 6

oblador:
Officially SB 6 doesn't support react-native, but seems to work just fine to me.


techeverri (approved):

examples/react-native/storybook/addons.js
```diff
@@ -1,4 +1,3 @@
-/* eslint-disable import/no-extraneous-dependencies, import/no-unresolved, import/extensions */
-
 import '@storybook/addon-actions/register';
 import '@storybook/addon-links/register';
+import '@storybook/addon-knobs/register';
```
techeverri, line 3: - `@storybook/addon-knobs` is not in the `package.json` as of today f7d5e65aee9b46787bcc7accd29a197dca55dc6f
- And `@storybook/addon-knobs` version `^6.1` (the one used for the addons) doesn't have a module called

techeverri: FYI @oblador

Issue #144, opened 2019-05-14
Is it possible to show a side by side diff?


Fernando-Abreu:
It would be interesting if it could show a side by side diff comparison like this:
![image](https://user-images.githubusercontent.com/4967214/57713735-6888d980-7649-11e9-9c08-0a164392e25c.png)

I think it makes easier to see what changed. Is it possible to see the diff like this?

andre-brdoch:
Hey, you could do [this](https://github.com/oblador/loki/issues/76) to generate a HTML report, which does include side by side comparisons 😄

PR #311, opened 2021-03-01
Feature/diff with reference

Muritavo:
This PR enables a feature requested by #144 as it's useful for when comparing complex UI changes.

## What does it change:
- Added a new property called includeReferenceOnDiff (default: **false**)
- When the property is enabled, the reference image and the current images are concatenated side by side along with the diff result.

### Make the property work on the other two diff engines, **if someone want to do it, fell free to contribute to this PR as well**

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -58,6 +58,10 @@ There are three currently available options to choose from when comparing images
 }
 ```
 
+### `includeReferenceOnDiff`
+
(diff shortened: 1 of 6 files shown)


Muritavo:
Adding here so there is documented:

For the GraphicsMagick library there is a method [montage](https://www.npmjs.com/package/gm#montage), that combines multiple images side by side.

As for the looks-same library, the createDiff **without diff property** can be used to manually write the file to disk, with the other references

Issue #323, opened 2021-07-07
DeprecationWarning: `setAddon` is deprecated and will be removed in Storybook 7.0.


vespasianvs:
Can't see an issue for this in the list already, so adding it here. the file integration-core/src/decorate-storybook uses 'setAddon'. This is deprecated and going to be removed in Storybook 7.

Instead, stories etc need to be written in Component Story Format.

Happy to try to look at this and put in a PR if you want.

techeverri:
Thanks for reporting the issue @vespasianvs 💯 

Component Story Format (CSF) is supported already 👇 
https://github.com/oblador/loki/blob/master/examples/react/src/stories/csf.stories.js

Would CSF support be enough?
Or Would it still require changes to remove `setAddon`? If so, feel free to create a pull request 👍

vespasianvs:
I think it needs setAddon removing entirely. I just get a DeprecationWarning for every storybook test when I run my jest tests against storybook using their addon-storyshots. I traced the deprecation warning and it comes from the file above.

vespasianvs:
May take me a week or so to get round to looking BTW - gotta get on with my paid work first :)

vespasianvs:
Had a look and the area that is affected is for the .add.skip() and .lokiSkip() and .add.async() and .lokiAsync() style functionality. I can try to hack around it and keep it, but they are already deprecated. We could just get rid of that whole notation now and get people to switch to the parameters notation.

```
{ loki: { skip: true } }
{ loki: { async: true } }
```

It looks like they been deprecated for a while now?

techeverri:
Support for Storybook <5 was deprecated (but not removed) in https://github.com/oblador/loki/releases/tag/v0.23.0

vespasianvs:
So is it time to remove it now? Then it removes the need for that entire decorate storybook file from what I can see?

oblador:
@vespasianvs Yeah, it's been marked as deprecated for a year now, so I think it's time.

vespasianvs:
Ah....wait...I thought the whole file was deprecated, but lokiAsync is still used? Is the format of

`Story.parameters = { loki: { async: true } }`

Already implemented? I can't see it anywhere

oblador:
Use [`@loki/create-async-callback`](https://loki.js.org/flaky-tests.html#asynchronous-stories) for that!

vespasianvs:
[...]

PR #327, opened 2021-07-22
Remove setAddon Issue #323

vespasianvs:
As per issue #323 - I have removed the setAddon - in fact as discussed I have removed the deprecated functions entirely which means that there are no storybook deprecation warnings any more when running storyshots in jest. 

I have tested this on my own React JS projects and it all works fine and both skip and async work fine using CSF. I am not using React Native in any projects so although I have updated the call in the integration for React Native, I cannot test it. Hoping I've managed to do all of this correctly. Let me know if you spot any issues.


vespasianvs:
Hi, any news on looking at this? We're wanting to upgrade storybook but when we do loki fails because of this.

techeverri:
Please update the branch and I will trigger the checks 👍

oblador:
Closed in https://github.com/oblador/loki/pull/345

PR #345, opened 2021-11-13
Remove setAddon

oblador:
Duplicate of #327, but author didn't allow edits from me, thus I cannot rebase and start tests.


techeverri (approved):

msammarco:
Any word when this can be integrated?

vespasianvs:
D'oh - sorry I didn't see your comment on #327 ! I normally try to keep up on them but github didn't notify this time for some reason. At least it's sorted!

edgarkhanzadyan:
@oblador @vespasianvs hello! First of all thanks for all your work to open source community! I was trying to use loki@0.29.0 but I would get storybook.map function undefined, so i was thinking that this PR could've broken it for react-native.

In loki@0.28.0 `Invalid hook call. Hooks can only be called inside of the body of a function component.`

So the latest working version for RN is 0.27.0

PR #353, opened 2021-12-27
Add support for `.nvmrc`

techeverri:
## Summary

This pull request introduces a `.nvmrc` (node version file) that helps to align the required Node.js version to be able to run, build, and test the project locally and with GitHub Actions.  

We will continue to support the latest minor of Node.js 12 (LTS) https://github.com/oblador/loki/pull/353#issuecomment-1001339392

🔗 https://nodejs.org/en/about/releases/
image


oblador:
I think it's a good change to use .nvmrc but I'm not sure I want to drop node 12 support just yet as it's still in maintenance LTS, and we don't have any dependencies requiring it - only one integration tests that doesn't have the version locked. 

I would suggest to either specify a version for the `yarn create react-app` command or bump node version for just the integration test. What do you think?

techeverri:
> I think it's a good change to use .nvmrc

I'll keep this pull request open and revert the changes that drop support for the old LTS.
Should we change the required version to the latest minor of Node.js 12? @oblador 

> [...] but I'm not sure I want to drop node 12 support just yet as it's still in maintenance LTS, and we don't have any dependencies requiring it - only one integration tests that doesn't have the version locked.
> 
> I would suggest to either specify a version for the `yarn create react-app` command or bump node version for just the integration test. What do you think?

I will bump the Node version for the integration test 👍

techeverri:
Fix for the CLI integration tests in GitHub Actions https://github.com/oblador/loki/pull/354

techeverri:
We should rebase this branch once https://github.com/oblador/loki/pull/354 is merged 👍

oblador (approved):

PR #352, opened 2021-12-22
Add Me Salva! into users

RicardoFredes:
Hi! My team at Me Salva! is using Loki to test on Storybook ([Mars](https://mars.mesalva.com/)).
And we are very happy with this solution!
We would like to be part of your user list. Thank you very much!

image

PR #355, opened 2021-12-28
Publish to GitHub Pages automatically using GitHub Actions

techeverri:
## Context

🔗 https://loki.js.org/

Changes to the website (e.g., https://github.com/oblador/loki/pull/352) are published manually, and the process requires @oblador credentials for GitHub

https://github.com/oblador/loki/blob/f7d5e65aee9b46787bcc7accd29a197dca55dc6f/package.json#L15

The website used to be published automatically by Travis. Using the following config 👇 

https://github.com/oblador/loki/blob/f7d5e65aee9b46787bcc7accd29a197dca55dc6f/.travis.yml#L38-L49

The last Travis build from a couple of years ago failed with the following error 👇 

🔗 https://travis-ci.org/github/oblador/loki/jobs/698966287

```
gh-token is invalid. Details: GET https://api.github.com/user: 401 - Bad credentials // See: https://developer.github.com/v3
Preparing deploy
failed to deploy
```

## Summary

This project uses GitHub Actions nowadays to automate workflows. This pull request aims to introduce a new workflow to automatically publish the website to GitHub Pages after a commit with changes in the `website` folder is pushed to `master` (e.g., a pull request is merged).

I propose using the action below to publish to GitHub Pages 👇 
🔗 https://github.com/marketplace/actions/github-pages-action

_Note: it's the one used in the [microsoft/TypeScript-Website](https://github.com/microsoft/TypeScript-Website) project_

## Next steps

- Delete https://github.com/oblador/loki/blob/master/.travis.yml
- Generate previews of changes to the website in pull requests
  - Perhaps we could make Vercel deployments of the build using the GitHub Action below 👇 
  🔗 https://github.com/marketplace/actions/vercel-action


.github/workflows/website.yml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          node-version-file: '.nvmrc'
+      - name: Install dependencies
+        run: yarn --frozen-lockfile --non-interactive --silent --ignore-scripts
+      - name: Build website
+        run: yarn workspace @loki/website build
+      - name: Deploy to GitHub Pages
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
techeverri, line 29: It seems we don't have to worry about generating the `GITHUB_TOKEN` according to this 👇 


> For newbies of GitHub Actions: Note that the GITHUB_TOKEN is NOT a personal access token. A GitHub Actions runner automatically creates a GITHUB_TOKEN secret to authenticate in your workflow. So, you can start to deploy immediately without any configuration.

🔗 https://github.com/marketplace/actions/github-pages-action

.github/workflows/website.yml
```diff
@@ -0,0 +1,30 @@
...
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'website/**'
+
+jobs:
+  publish:
```
techeverri, line 11: Another approach would be to have both a `build` and `deploy` jobs separately so that the build step can also be used for the preview job in pull requests. This would avoid duplicating the config to generate the build but also requires extra config to store and retrieve the artifacts using [actions/upload-artifact](https://github.com/actions/upload-artifact) and [actions/download-artifact](https://github.com/actions/download-artifact)

.github/workflows/website.yml
```diff
@@ -0,0 +1,30 @@
+name: Website
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'website/**'
```
techeverri, line 8: We could also run this workflow manually
🔗 https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow

```yml
on: workflow_dispatch
```

oblador:
This is great! What I would prefer is to publish the website when there is a new version released instead of when there is a change to the website so that the website doesn't contain documentation about unreleased changes. But I think at least before it was a bit difficult to make releases with github actions and lerna because manual triggers with custom parameters wasn't supported (type of change; patch/minor/major). Is that supported now?

techeverri:
> This is great! What I would prefer is to publish the website when there is a new version released instead of when there is a change to the website so that the website doesn't contain documentation about unreleased changes. [...]

This I've done it recently, to trigger stuff when a release is created.
_I will push the changes soon™_ 

> [...] But I think at least before it was a bit difficult to make releases with github actions and lerna because manual triggers with custom parameters wasn't supported (type of change; patch/minor/major). Is that supported now?

Does it have to support custom parameters? If so, What for? @oblador

.github/workflows/website.yml
```diff
@@ -0,0 +1,27 @@
+name: Website
+
+on:
+  release:
+    types: [created]
```
techeverri, line 5: This should trigger the workflow only when a new release is created @oblador 
More info: https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#release

oblador (approved):

oblador:
It seems like it wasn't published for some reason 🤔 https://github.com/oblador/loki/runs/5702951041?check_suite_focus=true

.github/workflows/website.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  publish:
+    if: ${{ github.ref == 'refs/heads/master' }}
```
oblador, line 9: Probably because of this check?

techeverri: I think I was trying to keep this behavior 👇 

https://github.com/oblador/loki/blob/4be00352e8554d7aff69c49f670fc247a9d22e6e/.travis.yml#L40

techeverri: I'll remove that since it doesn't make sense to require a branch here

techeverri:
While checking why the job would show "This check was skipped" I found that you can enable logs for the runners and the steps to debug the workflows more easily @oblador 

https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging

.github/workflows/website.yml
techeverri: Fixed by https://github.com/oblador/loki/pull/375

Issue #360, opened 2022-01-17
Update shelljs to 0.8.5 to mitigate the vulnerability GHSA-64g7-mvw6-v9qj


kapi-caz:
Hi!

loki-core uses shelljs in version 0.8.3 which has the vulnerability [GHSA-64g7-mvw6-v9qj](https://huntr.dev/bounties/50996581-c08e-4eed-a90e-c0bac082679c/). There are no braking changes mentioned in 0.8.4 and 0.8.5 [here](https://github.com/shelljs/shelljs/releases). I checked out the code, updated shelljs and run the unit tests and didn't found an issue. Due to missing permissions I couldn't create a PR.

Thanks for your work!

techeverri:
> [...] Due to missing permissions I couldn't create a PR.

You have to fork the repository first. It should work as expected after that.
Feel free to create a pull request 👍

kapi-caz:
> > [...] Due to missing permissions I couldn't create a PR.
> 
> You have to fork the repository first. It should work as expected after that. Feel free to create a pull request 👍

Thanks! It's my first contribution to any open source project, hence didn't thought about forking!

techeverri:
There is something I just realized reading the report below 👇 
🔗 https://huntr.dev/bounties/50996581-c08e-4eed-a90e-c0bac082679c/

> If ShellJS scripts running locally are using ShellJS exec function, local users on the filesystem can read the stdout of the running ShellJS process to disclose sensitive information present in the privileged process.

We don't use the `exec` function. We only use the `which` method.

https://github.com/oblador/loki/blob/f7d5e65aee9b46787bcc7accd29a197dca55dc6f/packages/core/src/dependency-detection.js#L15-L17

So Loki is not really affected by https://github.com/shelljs/shelljs/issues/1058 and the patch https://github.com/shelljs/shelljs/pull/1060 doesn't really change anything for Loki from a code perspective.

What do you think? @oblador

oblador:
Yeah, and also Loki's version is not locked to the vulnerable version. Simply update your project's lockfile, no change is really needed in Loki.

PR #361, opened 2022-01-18
Update shelljs to 0.8.5 in core (#360)

kapi-caz:
Hi!
I've update the dependency "shelljs" to mitigate a vulnerability (see: #360). [Shelljs](https://github.com/shelljs/shelljs/releases) did some bug fixes and updated a dependency. I ran the tests mentioned in the README.md and only the cli integration tests fails on my machine independently of my change, hence it's a local problem and I suppose it is not caused by my change (Invalid hook call [...])

Constructive feedback is welcome ;)

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -23,6 +23,6 @@
     "access": "public"
   },
   "dependencies": {
-    "shelljs": "^0.8.3"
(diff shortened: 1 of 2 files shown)


oblador:
This should be addressed in your project instead, see comment in the original issue 👍

PR #362, opened 2022-01-24
Story query params in configuration

PSTime:
This feature allows you to configure stories more deeply. For example, we can support themes (https://storybook.js.org/addons/storybook-addon-themes). You just need to pass query param into configuration. 
Example: 
` "chrome.docker.dark": {
        "target": "chrome.docker",
        "width": 1980,
        "height": 1000,
        "looks-same": {
          "ignoreCaret": true
        },
        "storyQueryParams": {"theme": "dark"}
      },`
      
And then implement a simple script in `.storybook/preview-head.html` to convert query params to HTML tag class.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -113,7 +113,7 @@ module.exports = {
 ## `configurations`
 
 | Name                                 | Type      | Description                                                                                                                                               | Targets      |
-| ------------------------------------ | --------- | --------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------ |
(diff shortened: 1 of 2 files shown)


PSTime:
@techeverri  @oblador please take a look)

oblador:
Would it make more sense to just enable forcing dark mode instead? Should be doable with [`setAutoDarkModeOverride`](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setAutoDarkModeOverride).

PR #365, opened 2022-02-14
Update react version in runner

euphocat:


oblador:
Fixed in https://github.com/oblador/loki/pull/373

PR #376, opened 2022-03-27
Enable manual trigger event for website workflow

techeverri:
Publish the website to GitHub Pages when a new release is created or the workflow is manually triggered.

### References

- https://docs.github.com/en/actions/managing-workflow-runs/manually-running-a-workflow
- https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_dispatch

### DIFF
--- a/.github/workflows/website.yml
+++ b/.github/workflows/website.yml
@@ -1,6 +1,7 @@
 name: Website
 
 on:
+  workflow_dispatch:
   release:
     types: [created]
 


.github/workflows/website.yml
```diff
@@ -1,6 +1,7 @@
 name: Website
 
 on:
+  workflow_dispatch:
```
oblador, line 4: Are you sure there should be a colon at the end?

techeverri: Let me double check 👀

techeverri: An alternative way to do it is to use a tilde `~` that apparently is valid for `null` in tag resolution according to the YAML spec.

https://yaml.org/spec/1.2.2/#1032-tag-resolution

```suggestion
  workflow_dispatch: ~
```

techeverri: I take it back. It does work as it is 💯 
_I tested in another repo_ ✅ 

The following lines 👇 

https://github.com/techeverri/techeverri-cli/blob/5daf89d0b3f8ccdc39339bec7398b0f41619987c/.github/workflows/ci.yml#L6-L7

Make the workflow to have a manual trigger in the user interface

https://github.com/techeverri/techeverri-cli/actions/workflows/ci.yml
image

techeverri: > Are you sure there should be a colon at the end?

Yes. Final answer 👍

techeverri: This syntax is also used in the official docs from GitHub for when the trigger doesn't have any config.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#using-data-from-the-triggering-workflow

image

PR #377, opened 2022-04-02
Upgrade chrome-launcher dependency

### LINKED ISSUE #369: Windows connect ECONNREFUSED
Trying to run tests on windows 10

`npx loki test`

```
loki test v0.28.1
 FAIL  chrome.app
       Starting
       connect ECONNREFUSED ::1:50499
```

configuration 

```
"loki": {
    "configurations": {
      "chrome.laptop": {
        "target": "chrome.app",
        "width": 1366,
        "height": 768,
        "deviceScaleFactor": 1,
        "mobile": false
      }
    }
  }
```

At the same time, everything works fine in the same repository on the mac os

Is there a solution? did not find similar issues

geigerzaehler:
This fixes an issue where chrome would fail to launch if `localhost` resolved to the IPv6 address `::1` on Node 17.

Fixes #369.

### DIFF
--- a/packages/target-chrome-app/package.json
+++ b/packages/target-chrome-app/package.json
@@ -21,7 +21,7 @@
   "main": "src/index.js",
   "dependencies": {
     "@loki/target-chrome-core": "^0.29.0",
-    "chrome-launcher": "^0.13.4",
+    "chrome-launcher": "^0.14.1",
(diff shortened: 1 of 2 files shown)


oblador (approved):
Had to bump the node version in lambci, so most likely this change breaks Node 8 support for lambdas, but I'm OK with that.

PR #325, opened 2021-07-19
Add configurationFilter per story by parameters

crysadrak:
This [issue](https://github.com/oblador/loki/issues/280) can be fixed by filter stories which doesn't match configurationFilter parameter defined in story:
```
Default.parameters = {
	loki: {
		configurationFilter: 'desktop' // regex string
	}
};
```

let me know if it is ok for you and where to put story parameters docs. Any tests for this feature needed?

Thx
CD

PR #387, opened 2022-04-29
Pass story parameters to `fileNameFormatter`

oblador:
Will allow using storybook provided `fileName` and user provided parameters for setting the baseline image file names.


techeverri:
Closes https://github.com/oblador/loki/pull/325

PR #388, opened 2022-05-04
Add option of specifying media features for chrome

rafal2228:
This feature adds support for emulating specific CSS media features (like preferred color scheme). The [features](https://chromedevtools.github.io/devtools-protocol/tot/Emulation/#method-setEmulatedMedia) parameter of `setEmulatedMedia` can be specified via configuration file for any chrome-based target.

It is inspired by the idea proposed by @oblador's idea to override what users can emulate in [this commment](https://github.com/oblador/loki/pull/362#issuecomment-1036584581). It also was mentioned in [the following discussion](https://github.com/oblador/loki/discussions/386).

The PR also updates the application in the `examples` folder, but at the current time the configuration of it seems to be broken due to:

- https://github.com/oblador/loki/blob/0e7acbdd976b5549ed7b68be7e63d03173066bc4/examples/react/src/stories/index.stories.js#L24
- https://github.com/oblador/loki/blob/0e7acbdd976b5549ed7b68be7e63d03173066bc4/examples/react/src/stories/index.stories.js#L36

loki specific API usage, that seems to be misconfigured/deprecated. @oblador @techeverri if you could please provide some guidance on how to fix these issues, I would be more than happy to include it in this PR (together with updated loki's reference images)


techeverri:
Reference images for the new stories are missing

rafal2228:
> Reference images for the new stories are missing

Yeah, I mentioned in the description that I wasn't able to generate them due to `lokiSkip` and `lokiAsync` usage inside of the stories file. Should I remove the usage and generate new reference images?

techeverri:
`lokiSkip` was deprecated by https://github.com/oblador/loki/pull/228, but it should still work. And https://github.com/oblador/loki/pull/263 fixed and issue that broke `lokiAsync`

What errors do you get?

oblador:
Yeah, these are deprecated APIs and I guess it's time to remove them from the tests too (I kept them to not break accidentally). They master though and looking at the test output, it complains about reference image not being present (you added new stories and configurations without adding baselines, so that is kinda expected). If you're unable to run the tests locally (try rebasing in that case) you can also download the baselines from the test run: https://github.com/oblador/loki/suites/6373156022/artifacts/234968222

rafal2228:
> They run fine on master though and looking at the test output, it complains about reference image not being present (you added new stories and configurations without adding baselines, so that is kinda expected). If you're unable to run the tests locally (try rebasing in that case) you can also download the baselines from the test run: https://github.com/oblador/loki/suites/6373156022/artifacts/234968222

Seems like the baseline is not actually correct. The tests run fine, but all examples from the above mentioned file (`index.stories.js`) are missing. I guess it's caused by the same problem with `lokiSkip` and `lokiAsync`.

> Yeah, these are deprecated APIs and I guess it's time to remove them from the tests too (I kept them to not break accidentally).

So should I delete the usage of those APIs from examples and update baseline images as part of this PR or do we need this work done on separate PR?

oblador:
I can do this in a separate PR 👍

oblador:
Fixed in #390!

rafal2228:
> Sorry for being annoying here, but don't you think it would enough to do this for just one test instead of adding two more configurations? See this example: https://github.com/oblador/loki/blob/master/examples/react/src/api/Decorators.stories.js#L56

This is actually a brilliant idea 👍 I had to update chrome target to allow passing of story parameters to `tabOptions`

oblador:
Ah, didn't know this wasn't supported yet :D Thanks for fixing/adding!

oblador (approved):

docs/configuration.md
```diff
@@ -112,17 +112,18 @@ module.exports = {
...
+| **`storiesFilter`**                  | _string_  | Same as `loki.storiesFilter`, but applied to only this configuration.                                                                                                                                                                                              | All          |
+| **`chromeSelector`**                 | _string_  | Same as `loki.chromeSelector`, but applied to only this configuration.                                                                                                                                                                                             | `chrome.*`   |
+| **`preset`**                         | _string_  | Predefined bundled configuration, possible values are `Retina Macbook Pro 15`, `Retina Macbook Pro 15 Dark Mode`, `iPhone 7`, `iPhone 7 Dark Mode`, `iPhone 5`, `iPhone 5 Dark Mode`, `Google Pixel`, `Google Pixel Dark Mode`, `A4 Paper`, and `US Letter Paper`. | `chrome.*`   |
```
techeverri, line 121: We should remove this from the docs before merging

techeverri (changes requested):

PR #390, opened 2022-05-10
Drop legacy APIs in example project

oblador:
This PR removes `.lokiSkip` and `.lokiAsync` stories as they are no longer supported.

While I was at it I also refactored most of the stories into CSF and separated them a bit more into sections based on what they test.


techeverri:
There is baseline that's missing 

```
FAIL chrome.docker/chrome.iphone7/Text/With External Font: No reference image found
```

https://github.com/oblador/loki/runs/6366856526?check_suite_focus=true
image

oblador:
Git and case-insensitive file systems never fail to disappoint 😢

Issue #234, opened 2020-04-23
Approve changes more files than the reported difference


donaldpipowitch:
It can happen that `approve` changes file which were not marked at different before. (E.g. they pass the test, but apparently something has changed.) It would be nice, if approve would only change files which were also listed as different in the `test` phase to avoid confusion ❤️

marcinczop:
Any update on this one?:)

techeverri:
`loki approve` is also meant to be used after `loki update` so it has to mark everything as accepted even files that were just created or with minor changes that don't trigger `loki test` to fail because of the `chromeTolerance`

techeverri:
But I can see how having a flag could help with your use case.
Something like `loki approve --diffOnly`

siyavash4812:
Has there been any work on this? At the moment if there is even one image in my diff folder, after running approve it will replace all the pictures in the reference folder

markbrockhoff:
I just implemented the `--diffOnly` flag described by @techeverri. It would be great if you or @oblador could have a look at it. 😃
#402

Issue #317, opened 2021-04-23
no images found to approve 


florian-milky:
`loki test` returns a bunch of failed tests
when I want to approve all changes with `loki approve`, it tells me `no images found to approve`
Why is that?

techeverri:
> `loki test` returns a bunch of failed tests
> when I want to approve all changes with `loki approve`, it tells me `no images found to **approve`**

The code that prints the error is very simple and the only obvious way is that there is no valid PNGs in the output folder

https://github.com/oblador/loki/blob/8beda16fcda4841c4e23c7353880e94498cbadcc/packages/runner/src/commands/approve/index.js#L15-L21

> Why is that?

Hard to know without a way to reproduce it.

florian-milky:
Thanks @techeverri  I'm gonna try to debug it :)

techeverri:
If you find a bug please reopen this issue or create a new one 👍

florian-milky:
@techeverri `fs.readdirSync(outputDir)` returns only top folders `desktop` and `mobile` which are my configurations

techeverri:
How does your `fileNameFormatter` look like?
🔗 https://loki.js.org/configuration.html#filenameformatter

Does it happen on Windows, Linux or macOS?

florian-milky:
> How does your fileNameFormatter look like?

```
fileNameFormatter: ({ configurationName, kind, story }) => {
    return `${configurationName}/${kind}--${story}`;
  },
```

I'm on macOS.

Is loki supposed to work well with subdirectories?

I have a script that generates the `approve` cmd file by file so I fixed that. However I noticed it's a bit confusing because in the story filter I had to remove the `--` that I get from the file name.

techeverri:
The `defaultFileNameFormatter` looks like this 👇 

https://github.com/oblador/loki/blob/8beda16fcda4841c4e23c7353880e94498cbadcc/packages/runner/src/commands/test/get-output-paths.js#L9-L10

It doesn't generate subfolders and so `fs.readdirSync(outputDir)` works as expected with a flat structure

techeverri:
It sounds like we need extra logic in all places we read from `outputDir` to support when the config has a custom `fileNameFormatter`

florian-milky:
> It sounds like we need extra logic in all places that read from outputDir to support when the config has a custom fileNameFormatter

Would you recommend keeping a flat structure? I don't think it's a deal breaker, you're not really supposed to browse the images manually anyway

[...]

PR #402, opened 2022-07-07
Add diff only option to approve command

markbrockhoff:
Closes: #234 

I implemented the `--diffOnly` option suggested by @techeverri in his [comment](https://github.com/oblador/loki/issues/234#issuecomment-788862248).
It allows the user to only approve changed images that are inside the difference directory. Therefore it is neccessary to run `loki test` beforehand.
The option can be used to reduce unneccessary updates of images that might have changed slightly but not enough to fail `loki test`.

It's even possible to set `diffOnly` to true inside the configuration to make this the default behaviour. In case the setting in the configuration should be overwritten the cli flag can be used.
Example: `loki approve --diffOnly false`

As this flag does not change the default behaviour of loki, I hope that it can be merged and released soon to fix the issues described inside #234


techeverri:
What do you think? @oblador

gidztech:
This option would be really useful, since we have some developers using M1 chips and others on Intel, and there are different visuals being output that don't reach our threshold. `loki approve` currently adds a ton of unnecessary changes to our PRs. For now, we're able to just use the generated `loki update` command that `loki test` provides on test failures, which includes a regex filter of stories. I guess `loki approve --diffOnly` will produce the same result as this?

We had to remove our custom `fileNameFormatter` to support this. I see that this is a known issue https://github.com/oblador/loki/issues/317

xurion:
Interested to find out if this can be merged or at least considered. the team I'm on with 15 other engineers face the issue of M1 chips vs. other hardware and it means we keep having to jump through unnecessary tasks to keep things clean.

Flirre:
Is this something that is being considered? This would reduce a lot of the problems my team have had with Loki.

josephluck:
We're running in to this as well. @oblador it'd be awesome to consider merging this improvement

oblador (approved):

oblador:
Good feature, thanks for the patience on this, I have my notifications muted so it few under my radar.

Issue #405, opened 2022-07-15
ECONNREFUSED error again on windows


bonesoul:
although i've commented here https://github.com/oblador/loki/issues/369 i wanted to create a new issue as the old one is already closed.

still getting the error with windows witn node 18.

```
λ loki update
loki update v0.30.3
 FAIL  chrome.app
       Fetching stories
       connect ECONNREFUSED ::1:55504
Some visual tests failed to run
```

any chances to test with chrome-launcher 0.15.1?

techeverri:
You can try updating `chrome-launcher` and testing on your Windows machine @bonesoul
Feel free to open a pull request with the dependency update, if that fixes it 👍

vladbulyukhin:
I also encountered this problem. It turned out that in my case, it was not `chrome-launcher`, but `chrome-remote-interface` that caused it. I created a PR (#418) with a newer version of `chrome-remote-interface`, but overriding the `chrome-remote-interface` version in your package.json also helps and can be used as a workaround.

**Workaround**
```
// package.json
{
  "overrides": {
    "chrome-remote-interface": "0.31.3"
  },
}
```

AiranBalor:
> I also encountered this problem. It turned out that in my case, it was not `chrome-launcher`, but `chrome-remote-interface` that caused it. I created a PR (#418) with a newer version of `chrome-remote-interface`, but overriding the `chrome-remote-interface` version in your package.json also helps and can be used as a workaround.
> 
> **Workaround**
> 
> ```
> // package.json
> {
>   "overrides": {
>     "chrome-remote-interface": "0.31.3"
>   },
> }
> ```

Did't work, still got same error

vladbulyukhin:
> > I also encountered this problem. It turned out that in my case, it was not `chrome-launcher`, but `chrome-remote-interface` that caused it. I created a PR (#418) with a newer version of `chrome-remote-interface`, but overriding the `chrome-remote-interface` version in your package.json also helps and can be used as a workaround.
> > **Workaround**
> > ```
> > // package.json
> > {
> >   "overrides": {
> >     "chrome-remote-interface": "0.31.3"
> >   },
> > }
> > ```
> 
> Did't work, still got same error

@AiranBalor that's strange. It worked for me, I use it as a workaround in my GitHub Action. Have you tried removing `node_modules` and `package-lock.json` and re-installing everything after adding "overrides" to your `package.json`?

AiranBalor:
[...]

PR #418, opened 2022-09-23
Update chrome-remote-interface to fix connection problem with Node 18

vladbulyukhin:
@techeverri

Closes: #405 

This PR fixes the `ECONNREFUSED` error when using `loki` with Node 17+ described in #405.

DNS lookup returns IPv6 with Node 17+, which breaks the `chrome-remote-interface` initialization. However, this problem was fixed in a newer version of the `chrome-remote-interface` package, so updating the package solved the issue for me.

### DIFF
--- a/packages/target-chrome-app/package.json
+++ b/packages/target-chrome-app/package.json
@@ -22,7 +22,7 @@
   "dependencies": {
     "@loki/target-chrome-core": "^0.30.1",
     "chrome-launcher": "^0.14.1",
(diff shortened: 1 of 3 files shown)


oblador:
Thanks for your PR. It seems yarn.lock is not updated, would you mind running yarn and committing it?

vladbulyukhin:
@oblador I updated `yarn.lock` and added it to the initial commit.

PR #422, opened 2022-11-07
add types for create-async-callback and loki-is-running

AntonNiklasson:


oblador:
Thanks for your PR. Don’t you have to add a reference to the types in the package.json too?

AntonNiklasson:
Sorry for not getting back here, I missed your reply completely. I'm not sure, I'll validate with our setup!

AntonNiklasson:
> Don’t you have to add a reference to the types in the package.json too?
You're right, I've updated the packages accordingly :)

oblador (approved):

Issue #71, opened 2018-06-19
Roadmap considerations


oblador:
Posting this for discussion would love any type of input here, either do/don't/doesn't matter, do this instead, other considerations, what's most important to you, what you'd be able to help out with etc. 

### Move to jest
It's hard to write a great test runner with all types of different outputs (like JUnit, coverage etc). Currently we're using Listr to get a slightly nicer output of the progress, but it's flawed because it will swallow `console.log` output, is limited in how it handles concurrency also sometimes it will just output every frame instead of updating it. So my thought was to not try to innovate/solve all of these problems but use something existing that works well and my thought was to use jest.

### Native concurrency
Running native tests are slow due to the screenshot process in the simulator itself being slow. That can be improved in multiple ways, for one we could take the screenshot from within the app itself and feed it over websockets (I've done this and it's fast) but it has the problem of not being accurate in some edge cases and requires integration of native libraries which makes it harder to get started. It should now be possible to run multiple simulators simultaneously but there's currently no way of sending commands to a specific connected device without doing alterations to the code itself. I've described it here: https://github.com/oblador/loki/issues/24 I guess my consideration is if I should give up on being a very light integration or trying to solve it in making storybook more flexible. 

### Visualisation/management
In my research I didn't find a good OSS library that _just_ handled the diff management well (showing you what the differences were in the browser and enabling you to accept changes). There are some projects, but they either try to solve a bigger problem that are similar but not really the same or they just aren't solving it well enough. There's a few hosted services that one could use, which is probably the best solution for most cases. However there are OSS projects w/o money and sensitive projects where using hosted solutions is difficult for multiple reasons where a self hosted thing would be much easier. At some point I wanted to make a stand alone project (I named it Sigyn) to cater to these needs, but I'm not sure if it's better to focus on solving loki problems first. I'm afraid of that both will hinder loki adoption, either loki itself doesn't get enough attention to be good or the overall experience is not good to be used by devs. 

Ping @nowells & @nuc

nowells:
> Move to jest

I like the idea of using `jest` to be runner, it seems like a great task runner with built in mechanics for concurrency, CLI outputting, and test reporting. I was looking over `Listr` and noticed some of these pain points where it was not really meant to solve the problems we are evolving into (but provided a nice system to start with for sure)

> Native concurrency

[...]

PR #427, opened 2022-12-13
Add support for targetting a specific device ID when running in React Native

levibuzolic:
Currently when running loki on a machine with multiple iOS simulators, loki will use the first booted device to use for screenshots.

In our CI we run 3 side-by-side iOS simulators on each machine which results in screenshots being captured from the incorrect simulator.

This PR adds a new `--device` flag to loki which allows you to specify the device ID to use for screenshots. We've [forked osnap](https://github.com/ferocia/osnap) which is no longer maintained and [added support for selecting a device ID for iOS](https://github.com/ferocia/osnap/pull/1). It already had support for Android.

I know there was some dicussion about proper concurrency support in #71, but given the age of that issue I assume there hasn't been any progress in that space.

### DIFF
--- a/packages/target-native-android-emulator/src/create-android-emulator-target.js
+++ b/packages/target-native-android-emulator/src/create-android-emulator-target.js
@@ -3,15 +3,15 @@ const tempy = require('tempy');
 const osnap = require('osnap/src/android');
 const { createWebsocketTarget } = require('@loki/target-native-core');
 
-const captureScreenshot = async () => {
+const captureScreenshot = async (device) => {
   const filename = tempy.file({ extension: 'png' });
-  await osnap.saveToFile({ filename });
+  await osnap.saveToFile({ filename, device });
   const screenshot = await fs.readFile(filename);
   await fs.unlink(filename);
   return screenshot;
 };
 
-const createAndroidEmulatorTarget = (socketUri) =>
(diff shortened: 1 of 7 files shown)


oblador:
Thank you for your PR! Yes I’d prefer to have proper concurrency support, but that would involve also having to deal with juggling simulators and probably building the app itself which is a big undertaking and would likely require more config and a big maintenance burden. I think this is a great compromise that doesn’t have any of those downsides.

packages/target-native-android-emulator/src/create-android-emulator-target.js
```diff
@@ -3,15 +3,15 @@ const tempy = require('tempy');
 const osnap = require('osnap/src/android');
 const { createWebsocketTarget } = require('@loki/target-native-core');
 
-const captureScreenshot = async () => {
+const captureScreenshot = async (device) => {
```
oblador, line 6: Hmm, I wonder if this approach actually works with concurrency. Wouldn’t the story change on all simulators at once? How else would you target it? Do you build and run three separate servers on different ports?

levibuzolic: Yeah, our setup runs 3 seperate packagers, instances of storybook and loki all running different randomised ports.

oblador: Okie 👍

PR #429, opened 2022-12-13
Update osnap target to @ferocia-oss/osnap

levibuzolic:
Sorry @oblador, I accidentally missed these changes from #427

Was still targetting the old deprecated `osnap` package.


oblador:
I noticed, trying to fix that [here too](https://github.com/oblador/loki/pull/428), but makes sense to it separately

oblador (approved):

PR #433, opened 2023-01-22
upgrade looks same to latest

Mgolbraikh:
I've seen that looks-same package was upgraded to 8, thought this might be nice as a first time commit.
Ive run "yarn test" and all passes.


Mgolbraikh:
Hi @oblador , First time trying to contribute, What do you think?

PR #450, opened 2023-04-06
Update @ferocia-oss/osnap to fix issues with ES Modules

levibuzolic:
`@ferocia-oss/osnap` received some sub-dependency updates in 1.3.1 which resolved some outstanding security issues and CVEs, unfortunately some of these updates included some packages which have moved exclusively to ES Modules which broke in Common JS environments.

This has been fixed as of 1.3.5, the broken versions have been deprecated on NPM, however to help other developers avoid ending up with a bas version I figured it'd be good to bump the required version of `@ferocia-oss/osnap` in the `package.json`

Fixes #448
Fixes #447

Issue #436, opened 2023-02-15
Cannot call extract() unless you call cacheAllCSFFiles() first


p3k:
I am getting this error when I run `loki test` with the [storyStoreV7 feature](https://storybook.js.org/docs/react/configure/overview#on-demand-story-loading) being enabled in the Storybook configuration.

There is a corresponding issue in Storybook’s issue tracker: https://github.com/nuxt-community/storybook/issues/374

It seems to need a fix in application code, e.g. https://github.com/reg-viz/storycap/issues/594

AleksRap:
Hi. I have the same problem. I tried to run loki using docker with image yukinying/chrome-headless-browser:90.0.4408.0

AleksRap:
I solved this problem like this:

Go to the @loki/browser library. Find a file get-stories.js . You need to make the getStories function async and add a line `await window.__STORYBOOK_CLIENT_API__.storyStore.cache all CSV files();` before returning

But my tests failed with error  "Timeout after 10000ms"

trunneml:
@AleksRap thanks your changes worked for me

pavmax:
Had the same issue, fixed with @AleksRap changes.
I think these changes should be integrated in the library.

HoneyQumo:
@AleksRap you are the best, it helped me a lot!

HoneyQumo:
@AleksRap How can this be used for CI?

ArthurGoupil:
Setting
```
  features: {
    storyStoreV7: false,
  }
```
in Storybook `main.js`
and `"chromeSelector": "#storybook-root > *",` in Loki config fixed everything for me, thanks @AleksRap !

oblador:
Fixed in 0.31.2 👍

PR #451, opened 2023-04-17
Remove support for Node 14, raise minimum version to 16

levibuzolic:
As discussed in #450, this raises the minimum node version support to 16 to fix some compatibility issues with the latest version of various dependencies.

https://github.com/oblador/loki/pull/450#discussion_r1167658000


oblador:
I suspect the failing integration test is due to the SB 7 release that somehow broke stuff. Does it pass on master for you?

.nvmrc
```diff
@@ -1 +1 @@
-v14.17.3
+v16.20.0
```
techeverri, line 1: We should also update this one 👇 

https://github.com/oblador/loki/blob/7ca6aa0101151dd7ec56652708cdafd1083079cf/docs/serverless.md?plain=1#L66

techeverri: I'd also suggest adding " (Node.js v16.x or newer)" to the "prerequisites" sections

https://github.com/oblador/loki/blob/7ca6aa0101151dd7ec56652708cdafd1083079cf/packages/loki/README.md?plain=1#L25

https://github.com/oblador/loki/blob/7ca6aa0101151dd7ec56652708cdafd1083079cf/docs/getting-started.md?plain=1#L8

package.json
```diff
@@ -15,7 +15,7 @@
     "publish-docs": "GIT_USER=oblador yarn workspace @loki/website publish-gh-pages"
   },
   "engines": {
-    "node": ">=14.0.0"
+    "node": ">=16.0.0"
```
techeverri, line 18:
```suggestion
    "node": ">=16"
```

techeverri: I used https://semver.npmjs.com/ to compare the matches
_Package: `node`, Version range: `>=16`_

techeverri: Shouldn't we have at least the same version that we have in the `.nvmrc`?

```suggestion
    "node": ">=16.20.0"
```

levibuzolic: @techeverri the `.nvmrc` is for development, but the `engines` is for consumers. There's no strict reason they need to be identical.

levibuzolic:
> @oblador I suspect the failing integration test is due to the SB 7 release that somehow broke stuff. Does it pass on master for you?

`master` doesn't pass locally anymore because Storybook 7 also requires Node 16 now.

levibuzolic:
@oblador Digging into the failure a little bit more, it looks Storybook 7 requires a web server because they've switched to usedusing JavaScript modules (`.mjs`) which can't be loaded from the local filesystem.

I quickly hacked in a `http-server` wrapper around the static build (though we might as well just run the storybook server directly) which gets a little further, however it looks like loki [isn't compatible with Storybook 7](https://github.com/oblador/loki/issues/436), so that'll need to be fixed first.

oblador:
@levibuzolic 0.31.2 is out now with storybook 7 fixes, could you please give it another try? 🙏

levibuzolic:
@oblador should be good to go now.

oblador (approved):

oblador:
Thanks!

PR #459, opened 2023-05-26
add host-gateway only when docker server is >= v20

jafern14:


jafern14:
I still need to add a test to mock out the docker cli `version` cmd to prove that it works. I wanted to get your thoughts on this before doing so though. Thanks!

oblador:
I'm cool with the change! Make sure to update the yarn.lock by running `yarn` in the root.

oblador:
It seems this causes issues in GH Actions: `Error: listen EACCES: permission denied 0.0.0.0:491`

oblador:
Merging with master again and we can see if this is just flake :-)

oblador:
Different error now that seems related: `1 request failed to load; http://host.docker.internal:4567/1000/https://loki.js.org/img/favicon.png`

jafern14:
Sounds good to me and thanks for following up on this! I appreciate it 👍. 

I don't have time to look into it right now, but I'm wondering if there's some issue with the CI env and the port that docker is using behind the special `host-gateway`.

Do you know what version of docker is being used in the CI env?

oblador:
Looks like 24: https://github.com/actions/runner-images/blob/ubuntu22/20231025.1/images/linux/Ubuntu2204-Readme.md#tools

oblador:
Btw, regarding testing, I think it might be better to just have a test matrix for the integration tests rather than mocking stuff.

Issue #473, opened 2023-08-23
Failed to load stories in chrome.docker, loki version 0.32.0


jamiepenney:
While running this command to update my local reference images: `yarn build:storybook --quiet && loki update --requireReference --reactUri file:./storybook-static --chromeDockerImage yukinying/chrome-headless-browser-stable:latest`

I'm getting the following error:

```
chrome.docker
       Fetching stories
       Failed fetching stories because the server is down
       Try starting it with "yarn storybook" or pass the --port or --host arguments if it's not running at http://192.168.1.131:49399
```

I have nailed this down to the chrome docker image not being able to access Storybook running on my host machine. I'm running this on an M1 Macbook running Docker Desktop 4.22.0, and if I enter the docker image while it is running then I can see that it is unable to access the host by running `wget http://192.168.1.131:49399`. If I access that URL locally, Storybook is successfully running on it.

If I manually edit the IP detection code on this line https://github.com/oblador/loki/blob/master/packages/target-chrome-docker/src/create-chrome-docker-target.js#L70 to instead always use `host.docker.internal` as the IP, then everything works correctly.

Is there a reason not to use `host.docker.internal` as the local IP address by default? An earlier commit added the docker argument required to ensure this works correctly on Linux: `--add-host=host.docker.internal:host-gateway`.

I'm not sure of the broader implications of this change. There doesn't appear to be any way to override the IP address when you are using the static Storybook build option, but I could add one if you think it's too risky to use `host.docker.internal` by default.

huyit88:
Yes, i got the same problem . Can you @jamiepenney  elaborate your workaround on mac? 
What do you change to make it work? 
Thanks!

jamiepenney:
@huyit88 I've made a draft PR with my suggested changes

huyit88:
Thank you.
It works. 
I am not sure why it's relative to host issue. Be the way, With different network provider, the results are also different. 
I able to run loki test at my company but get the same error when i run it at home and can't figure it out why. 
Not sure if this workaround can apply for below issue. 
https://github.com/oblador/loki/issues/142
Anyway, thank for your input.

fttneto:
I'm trying for a while to run Loki with `chrome.docker` settings and the only way I found to get it working is by replacing the code suggested in the PR opened by @jamiepenney and also running it through build and passing `--reactUri` option.

Trying to run it using `yarn storybook` will make the process fail with the same error that @jamiepenney said here.

[...]

PR #474, opened 2023-08-23
Use host.docker.internal instead of local IP address in chrome.docker

jamiepenney:
Resolves https://github.com/oblador/loki/issues/473

Fixes an issue where the Chrome docker container cannot access Storybook running on the host machine via IP address, instead using the Docker `host.docker.internal` DNS entry to get access to it.

I'm not 100% convinced this is the right approach, but I wanted to put it out there as a solution to my problem. Is there a scenario where using `host.docker.internal` for accessing the host from the docker container _won't_ work?

### DIFF
--- a/packages/target-chrome-docker/src/create-chrome-docker-target.js
+++ b/packages/target-chrome-docker/src/create-chrome-docker-target.js
@@ -66,20 +66,19 @@ function createChromeDockerTarget({
   }
   runArgs.push('--add-host=host.docker.internal:host-gateway');
 
-  if (dockerUrl.indexOf('http://localhost') === 0 || isLocalFile) {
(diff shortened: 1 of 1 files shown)


patrik909:
This also works for me on 0.34.0! Would be great to have this merged :)

sayonar2k:
it actually works, ty (:

Vorbert-Kruk:
Same here, seems to be resolving the issue.
Would be great to see it merged and released!

Vorbert-Kruk:
Hey @techeverri, @oblador
In a convenient time, could you review this PR?

Pushing it forward would resolve CI issues with which people are currently struggling.
There are few issues related to it, but the most relevant probably would be #473.

PR #476, opened 2023-09-14
chore: update storybook/react peer dep

davebrewer:
Resolves peer dep issue when using Loki with "@storybook/react" 7


squidjam (approved):
I see the code change will sort out issues for non-yarn users.

PR #478, opened 2023-10-11
Update target-chrome-docker's runningInDocker check for cgroupsv2

remleduff:
This fixes docker-in-docker for me, particularly in gitlab. Without this, it tries to connect to 127.0.0.1 rather than the container's correct IP address and fails with a timeout.


remleduff:
My biggest question is why we need this check at all. 

Doing the docker inspect unconditionally seems to work for me, and using the container's IP address should always be better I'd think?

PR #481, opened 2023-10-22
Use `amazon/aws-lambda-nodejs:16` Docker image

techeverri:

PR #483, opened 2023-10-31
Temporarily skip lambda tests

oblador:
Working on a real solution in #481, but skipping them for now until we manage to fix the runtime.


techeverri (approved):

oblador:
Closing this due to false premise; the lambda tests are passing on master, but chrome docker aren't.

PR #484, opened 2023-10-31
Lock Chrome version in docker tests

oblador:
Previously the unit tests used `latest` which caused them to suddenly fail because Chrome is crashing. Long term we would probably want to investigate why latest version is crashing (if it's an issue with the docker image or with the GitHub Action environment).

### DIFF
--- a/packages/target-chrome-docker/src/create-chrome-docker-target.js
+++ b/packages/target-chrome-docker/src/create-chrome-docker-target.js
@@ -43,7 +43,7 @@ const waitOnCDPAvailable = (host, port) =>
 
 function createChromeDockerTarget({
   baseUrl = 'http://localhost:6006',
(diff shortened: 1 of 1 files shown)


packages/target-chrome-docker/src/create-chrome-docker-target.js
```diff
@@ -43,7 +43,7 @@ const waitOnCDPAvailable = (host, port) =>
 
 function createChromeDockerTarget({
   baseUrl = 'http://localhost:6006',
-  chromeDockerImage = 'yukinying/chrome-headless-browser-stable',
+  chromeDockerImage = 'yukinying/chrome-headless-browser-stable:100.0.4896.127',
```
techeverri, line 46: 👍

techeverri (approved):

PR #486, opened 2023-11-02
Bump default chrome docker version

### LINKED ISSUE #434: `chrome.docker` differences when running in CI compared to on MacOS
Hi,

I would expect `chrome.docker` to have the same rendering on any machine, however we are noticing really tiny differences when running in CI:

diff

This is a canvas rendered graph, but again in docker I would expect everything to be the same.

Are there any options I can specify to improve this?

oblador:
Running latest version of chrome on GitHub Actions failed due to seccomp restrictions, and while our seccomp white listing was very extensive it didn't capture everything, so this PR simply allows everything to avoid it randomly breaking in the future (you can always opt out of this behavior with `--chromeDockerWithoutSeccomp`). 

Related: #484

Fixes #434


techeverri (approved):

techeverri:
How did you find out it was the seccomp restrictions? @oblador

oblador:
Added a bunch of logs is all :-)

Issue #493, opened 2023-12-14
Update wait-on to the latest version to fix CVE-2023-45857 vulnerability


gblessy:
wait-on version 5 uses vulnerable version of axios. The latest version of the wait-on uses axios 1.6, where the vulnerability have been fixed. 

we need to update wait-on from 5.4 to 7.2

The wait-on version is set here:
https://github.com/oblador/loki/blob/master/packages/target-chrome-docker/package.json

PR #495, opened 2023-12-21
Update wait-on dependency

felabr:
Update the wait-on dependency to ^7.2.0.
Earlier versions use an older version of axios with a security issue.
This should solve https://github.com/oblador/loki/issues/493

### DIFF
--- a/packages/target-chrome-docker/package.json
+++ b/packages/target-chrome-docker/package.json
@@ -27,7 +27,7 @@
     "execa": "^5.0.0",
     "find-free-port-sync": "^1.0.0",
     "fs-extra": "^9.1.0",
-    "wait-on": "^5.2.1"
+    "wait-on": "^7.2.0"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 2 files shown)


felabr:
Hi @oblador, not sure how things proceed from here on ... could you have a look at this tiny PR?

gatsbimantico:
Hello @felabr I was aiming to propose the same update.

I think loki is still a commonjs project, while wait-on or axios have moved into ESM and that's making your test fail.

In `loki/packages/target-chrome-docker/src/create-chrome-docker-target.js` it will need to use a dynamic import, instead of require to import `wait-on`. Luckily for us it's already used within a promise, so the asyncronous nature of the dynamic import shouldn't affect anything else.

Instead of
```
waitOn = require('wait-on');

Promise(...
  waitOn(...)
```
needs to be something like
```
Promise(...
  import('wait-on').then(module => module.default).then(waitOn => waitOn(...))
```

From the docs: https://nodejs.org/api/esm.html#import-expressions
> Dynamic import() is supported in both CommonJS and ES modules. In CommonJS modules it can be used to load ES modules.

Let me know your thoughts.

felabr:
Yes, that would be great!
I tried a lot, Promises and async/await, but didn't get it to work (yet). Whenever there's a dynamic import for `wait-on` it  keeps running for ever. It even stops working when I just add this line:
```
import('wait-on').then((module) => console.log(module));
```
and change the test command in `package.json` like so:
```
"test": "yarn node --experimental-vm-modules $(yarn bin jest)",
```

The leanest thing I came up with (in my own opinion) is this:
```
const waitOnCDPAvailable = async (host, port) => {
  const { default: waitOn } = await import('wait-on');
  await waitOn({.....});
};
```

Anyway, if anybody has an idea how to make the `wait-on` ESM and the dynamic import work here, please let me know.

flieks:
I also have this problem with loki and people that install our npm package are getting the vulnerability warning for that reason. Would be great if this can be fixed

oblador:
Fixed in [0.34.0](https://github.com/oblador/loki/releases/tag/v0.34.0)

PR #500, opened 2024-02-15
fix: make is-loki-running include types in the package

igorbej:
This PR fixes the issue where the type definitions are missing when installing the package from npm.

### DIFF
--- a/packages/is-loki-running/package.json
+++ b/packages/is-loki-running/package.json
@@ -17,7 +17,7 @@
   "license": "MIT",
   "files": [
     "src",
-    "*.d.ts",
+    "*.d.ts"
   ],
   "main": "src/index.web.js",
   "react-native": "src/index.native.js",


packages/is-loki-running/package.json
```diff
@@ -16,7 +16,8 @@
   },
   "license": "MIT",
   "files": [
-    "src"
+    "src",
+    "*.d.ts",
```
techeverri, line 20:
```suggestion
    "*.d.ts"
```

packages/is-loki-running/package.json
```diff
@@ -16,7 +16,8 @@
   },
   "license": "MIT",
   "files": [
-    "src"
+    "src",
+    "*.d.ts"
```
techeverri, line 20: Should we also add `!*.d.ts` to the `.npmignore`? @oblador 

https://github.com/oblador/loki/blob/2d833f03b6f2695e0a5b6e5f30e7b21e5e924373/.npmignore#L1-L6

oblador: Kinda weird to have both, I think that since the migration to a monorepo this file would be ignored

oblador (approved):

PR #519, opened 2024-06-05
bump react, react-dom and ink versions to latest

DekraIT:
**Description**
This pull request updates the dependencies for react, react-dom, and ink to their latest versions.

**Motivation**
The update is necessary to resolve peer dependency issues encountered when using pnpm with React version 18.3.1. This ensures compatibility and smooth installation for projects utilizing the latest React version.

**Context**
pnpm enforces strict peer dependency resolution. If a project uses React 18 but one of its dependencies requires React 17, pnpm will flag this as a conflict, potentially preventing successful installation. By aligning the library's dependencies with the latest React version, this PR mitigates such issues.

The ink package needs to be updated as well, as it relies on an older version of react-reconcile, which is also problematic for dependency resolution with the latest React version.


oblador:
The bump of Ink breaks CJS support which is why I've held off upgrading it.

PR #532, opened 2025-01-28
Fix approve command with custom fileNameFormatter

d8corp:


vestrelatlassian:
Hey, thanks for this! I tried out the changes locally and it seems to work. However, other folders within the reference folder is getting deleted.

Here's my `loki.config.js`,

```js
module.exports = {
  configurations: {
    'ios.iphone16': {
      target: 'ios.simulator',
    },
    'android.pixel6a': {
      target: 'android.emulator',
    },
  },
  fileNameFormatter: ({ configurationName, kind, story, parameters }) =>
    `${configurationName}/${kind}_${story}`,
};

```

Running `yarn loki update` produces reference images in the following file hierarchy,

- `.loki/reference/`
  - `ios.iphone16/`
  - `android.pixel6a/`

If I make some changes such that running `yarn loki test` will fail (ios first), running `yarn loki approve` updates the `ios.iphone16/` correctly BUT the `android.pixel6a/` gets deleted.

vestrel00:
Hey, thanks for this! I tried out the changes locally and it seems to work. However, other folders within the reference folder is getting deleted.

Here's my `loki.config.js`,

```js
module.exports = {
  configurations: {
    'ios.iphone16': {
      target: 'ios.simulator',
    },
    'android.pixel6a': {
      target: 'android.emulator',
    },
  },
  fileNameFormatter: ({ configurationName, kind, story, parameters }) =>
    `${configurationName}/${kind}_${story}`,
};

```

Running `yarn loki update` produces reference images in the following file hierarchy,

- `.loki/reference/`
  - `ios.iphone16/`
  - `android.pixel6a/`

If I make some changes such that running `yarn loki test` will fail (ios first), running `yarn loki approve` updates the `ios.iphone16/` correctly BUT the `android.pixel6a/` gets deleted.

-------

**UPDATE:** 

It seems like this behavior is unrelated to the changes in this PR. The "issue" is caused by this block of existing code;

```
fs.emptyDirSync(referenceDir);
fs.ensureDirSync(referenceDir);

files.forEach((file) =>
  fs.moveSync(path.join(outputDir, file), path.join(referenceDir, file))
);
```

In the case of the `fileNameFormatter` that I'm using, I should be emptying and moving files into the dir with `${configurationName}` within the `referenceDir` instead of the `referenceDir` itself. 

--------

**UPDATE 2:**

I'm able to workaround this "issue" by specifying the output and reference dirs; `yarn loki approve --output="./.loki/current/ios.iphone16/" --reference="./.loki/reference/ios.iphone16/"`

--------

**UPDATE 3:**

Specifying the dirs every time is a bit cumbersome. So, I added a new required cli arg `--configurationName` to ensure that the approve script only applies to a single configuration (i.e. `ios.iphone16` or `android.pixel6a`).

The command I'm using is now `yarn loki approve --configurationName="ios.iphone16"`.

Here's the code I'm using in case others want to use them; [code.txt](https://github.com/user-attachments/files/19558297/code.txt)

The code changes I made are rudimentary. Nothing fancy or proper. Just something that works for my setup. Ideally, the approve script would work without having to specify additional args such as the `--configurationName` that I added.

PR #547, opened 2025-07-30
Fix: Handle storybook V8 fetch call

bassem-i:
We are currently migrating our Storybook version from 7 to 8. We've noticed that:

A `fetch` call in Storybook V8 breaks loki's aws-lambda-renderer.

Storybook V8 initialization calls `getStoryIndexFromServer` [function](https://github.com/storybookjs/storybook/blob/4a5d3f5a9f497d2a30e5394c6bd4868cd2f87a21/code/core/src/preview-api/modules/preview-web/Preview.tsx#L193) which tries to load a file called `index.json` (A json with all the stories loaded in the storybook build) using `fetch`. This breaks Loki's aws-lambda-renderer because it depends on file-based `file://TMP_FOLDER/iframe.html?id=STORY_ID` access. (File-based access doesn't support `fetch`).

This PR fixes the `failed to fetch` error by handling the fetch call without actually fetching. 

1. Loki's aws-lambda-renderer script (running on lambda) does a `readFileSync` to load the content of `index.json` file.
2. Loki's aws-lambda-renderer script injects a script into the chrome tab running Storybook to overwrite the `fetch` function implementation as follows:

* If fetch is called with `index.json`, it returns the stories list. 
* If fetch is called with any other input, it invokes the original fetch.

### DIFF
--- a/packages/browser/src/handle-stories-fetch.js
+++ b/packages/browser/src/handle-stories-fetch.js
@@ -1,0 +1,39 @@
+/**
+ * Storybook V8 initilization includes a fetch call to the index.json file. The index.json file includes a json object
+ * with all the stories from the storybook build.
+ *
+ * Since Loki is accessing the storybook build using file-based urls (file://tmp/storybook-static/iframe.html),
+ * Chromium will not be able to fetch the index.json file. (Fetch calls are blocked for security reasons.)
+ *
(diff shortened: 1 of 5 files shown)


bassem-i:
I've found another approach that keeps the file-based url access: https://github.com/oblador/loki/pull/546

